PR #4, opened 2023-11-16
jammy instead of bionic

munakoiso:


usernamedt:
Fixed in https://github.com/yandex/pgconsul/pull/5

PR #24, opened 2024-07-12
Verify timeline before lock acquire

munakoiso:
Somehow we can have a cluster with two primary hosts (<=1 is open) [A - with lower timeline and B - with greater timeline], and enabled maintenance. 

Right after disabling maintenance, host A can take the leader_lock faster, than the preferred the primary host - B. 
Host B will try to return to cluster, but it doesn't make any sense. Unfortunately, host B will wait for the recovery timeout - connecting to host A is impossible because it is closed by the local pgconsul. Therefore, the cluster will be unavailable until the recovery timeout has passed.

That is why we must verify the timeline before acquiring the leader lock.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -382,6 +382,12 @@ class pgconsul(object):
                 self.resolve_zk_primary_lock(my_hostname)
                 return None
 
+            # We shouldn't try to acquire leader lock if our current timeline is incorrect
+            if self.zk.get_current_lock_holder() is None:
+                # Make sure local timeline corresponds to that of the cluster.
+                if not self._verify_timeline(db_state, zk_state, just_check=True):
+                    return None
+
             if not self.zk.try_acquire_lock():
                 self.resolve_zk_primary_lock(my_hostname)
                 return None
@@ -894,7 +900,7 @@ class pgconsul(object):
         else:
             self._is_single_node = self.zk.exists_path(self.zk.SINGLE_NODE_PATH)
 
-    def _verify_timeline(self, db_state, zk_state):
+    def _verify_timeline(self, db_state, zk_state, just_check=False):
         """
         Make sure current timeline corresponds to the rest of the cluster (@ZK).
         Save timeline and some related info into zk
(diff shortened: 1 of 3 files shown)


tests/features/pgconsul_util.feature
```diff
@@ -523,7 +523,7 @@ Feature: Check pgconsul-util features
         When we release lock "/pgconsul/postgresql/alive/pgconsul_postgresql1_1.pgconsul_pgconsul_net" in <lock_type> "<lock_host>"
         And we release lock "/pgconsul/postgresql/alive/pgconsul_postgresql2_1.pgconsul_pgconsul_net" in <lock_type> "<lock_host>"
         And we release lock "/pgconsul/postgresql/alive/pgconsul_postgresql3_1.pgconsul_pgconsul_net" in <lock_type> "<lock_host>"
-        And we release lock "/pgconsul/postgresql/leader" in <lock_type> "<lock_host>" with value "pgconsul_postgresql1_1.pgconsul_pgconsul_net"
```
munakoiso, line 526: There is no step like "we release lock.*with value.*"

tests/features/maintenance.feature
```diff
@@ -23,8 +23,10 @@ Feature: Check maintenance mode
         """
         Then <lock_type> "<lock_host>" has holder "pgconsul_postgresql1_1.pgconsul_pgconsul_net" for lock "/pgconsul/postgresql/leader"
         When we set value "enable" for key "/pgconsul/postgresql/maintenance" in <lock_type> "<lock_host>"
+        And we wait "10.0" seconds
```
munakoiso, line 26: Here and below we should wait til pgconsul's at all hosts read the maintenance node

EinKrebs (approved):

src/main.py
```diff
@@ -909,14 +915,14 @@ def _verify_timeline(self, db_state, zk_state):
...
                 # Actually we should never get here but checking it just in case.
                 # Here we should end iteration and check and probably close primary
                 # at the begin of primary_iter
                 return None
         # If ZK does not have timeline info, write it.
-        elif zk_state[self.zk.TIMELINE_INFO_PATH] is None:
+        elif zk_state[self.zk.TIMELINE_INFO_PATH] is None and not just_check:
             logging.warning('Could not get timeline from ZK. Saving it.')
             self.zk.write(self.zk.TIMELINE_INFO_PATH, db_state['timeline'])
         # If there is a mismatch in timeline:
```
vicpopov, line 928: You can end up with `tli_res is False` (https://github.com/yandex/pgconsul/pull/24/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R934) while just_check is True.
In that case you will be doing modifications in Zookeeper while.
That is wrong because you don't have a lock and information that noone has it can be stale.

PR #28, opened 2024-07-18
Fix timeline verification before acquiring leader lock

munakoiso:
https://github.com/yandex/pgconsul/pull/24/files#r1679448285

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -385,7 +385,7 @@ class pgconsul(object):
             # We shouldn't try to acquire leader lock if our current timeline is incorrect
             if self.zk.get_current_lock_holder() is None:
                 # Make sure local timeline corresponds to that of the cluster.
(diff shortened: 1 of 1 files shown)


EinKrebs (approved):

src/main.py
```diff
@@ -915,14 +915,16 @@ def _verify_timeline(self, db_state, zk_state, just_check=False):
         # If it does, but there is no info on replicas,
         # close local PG instance.
         if tli_res:
-            if not just_check and zk_state['replics_info_written'] is False:
+            if zk_state.get('replics_info_written') is False:
```
munakoiso, line 918: In case of not defined zk_state['replics_info_written'] we should return True

EinKrebs (dismissed):

EinKrebs (approved):

PR #30, opened 2024-07-22
Better managing of replication slots

munakoiso:


src/main.py
```diff
@@ -1083,24 +1089,50 @@ def _attach_to_primary(self, new_primary, limit):
...
-                logging.debug(hosts)
-                if not self.db.replication_slots('drop', hosts):
-                    logging.warning('Could not drop replication slots. Do not forget to do it manually!')
-            else:
-                logging.warning(
-                    'Could not get all hosts list from ZK. '
-                    'Replication slots should be dropped but we '
-                    'are unable to do it. Skipping it.'
-                )
+        if not need_slots:
```
EinKrebs, line 1094: need_slots can be inlined

EinKrebs (changes requested):

tests/steps/database.py
```diff
@@ -111,13 +112,16 @@ def get_config_option(self, option):
...
+        try:
+            self.cursor.execute(
+                """
+                SELECT pg_create_physical_replication_slot(%(name)s)
+            """,
+                {'name': slot_name},
+            )
+            return deepcopy(self.cursor.fetchone())
+        except DuplicateObject:
+            return True
```
vicpopov, line 124: Maybe return `[True]` to mimc structure of one row returned by `.fetchone()`?

src/pg.py
```diff
@@ -748,7 +748,7 @@ def replication_slots(self, action, slots):
                 if not self._create_replication_slot(slot):
                     return False
             else:
-                if current and slot not in current:
+                if current is not None and slot not in current:
```
vicpopov, line 751: `current` is never `None`, it can only be an empty list: https://github.com/yandex/pgconsul/pull/30/files#diff-1dd21d0846b6c985aaf231799772c4797a69429020c8d3781caf8e285d35db43R148

src/zk.py
```diff
@@ -188,11 +189,17 @@ def _write(self, path, data, need_lock=True):
             logging.error(event.exception)
         return not event.exception
 
-    def _init_lock(self, name):
+    def _init_lock(self, name, read_lock=False):
         path = self._path_prefix + name
-        self._locks[name] = self._zk.Lock(path, helpers.get_hostname())
+        if read_lock:
+            lock = self._zk.ReadLock(path, helpers.get_hostname())
```
vicpopov, line 195: According to documentation https://kazoo.readthedocs.io/en/latest/api/recipe/lock.html#kazoo.recipe.lock.ReadLock, it works together with `WriteLock`, but I don't see any in this PR.
`ReadLock` will never lock with other `ReadLock`s, I don't see any point in using it.

vicpopov (changes requested):

src/zk.py
munakoiso: ReadLock looks like an easy way for me to store many locks on one ZK node. Thus, we can read from only one node to find all the hosts that require replication slots on the current host.

src/pg.py
munakoiso: https://github.com/yandex/pgconsul/blob/71f48d1fb44716daa1e9e0aa76d136f55abc0bb7/src/pg.py#L145

current is None if we catch an error
Actually, i do not remember specific error, but i saw it in feature tests before i write this line

tests/steps/database.py
munakoiso: ok, but it's does not matter.
there is only one call of this function
https://github.com/yandex/pgconsul/blob/main/tests/steps/cluster.py#L331

src/main.py
```diff
@@ -1083,24 +1089,49 @@ def _attach_to_primary(self, new_primary, limit):
...
+            logging.warning(
+                'Could not get all hosts list from ZK.'
+                'Can not handle replication slots. We will skip it this time'
+            )
+            return
+        hosts_except_lock_holders = list(set(all_hosts) - set(slot_lock_holders))
+        non_holders_hosts = []
+        for host in hosts_except_lock_holders:
+            if host not in self._slot_drop_countdown:
+                self._slot_drop_countdown[host] = self.config.getfloat('global', 'drop_slot_countdown')
```
vicpopov, line 1116:
```suggestion
                self._slot_drop_countdown[host] = self.config.getint('global', 'drop_slot_countdown')
```
We won't be able to use this value as float, we don't have non-integer step while operating this countdown timers.

src/main.py
```diff
@@ -1083,24 +1089,49 @@ def _attach_to_primary(self, new_primary, limit):
...
+            self._slot_drop_countdown[host] -= 1
+            if self._slot_drop_countdown[host] <= 0:
+                non_holders_hosts.append(host)
+        for host in slot_lock_holders:
```
vicpopov, line 1120: I would suggest a single loop over `all_hosts` with single `if host in slot_lock_holders`.
It would be overall shorter and clearer.

src/main.py
```diff
@@ -1083,24 +1089,49 @@ def _attach_to_primary(self, new_primary, limit):
...
+            self._slot_drop_countdown[host] = self.config.getfloat('global', 'drop_slot_countdown')
+
+        # create slots
+        slot_names = [i.replace('.', '_').replace('-', '_') for i in slot_lock_holders]
```
vicpopov, line 1124: Use `helpers.app_name_from_fqdn`?

src/main.py
```diff
@@ -1110,6 +1141,25 @@ def _get_db_state(self):
         logging.info('Database cluster state is: %s' % state)
         return state
 
+    def _acquire_replication_source_lock(self, primary):
+        if not self.config.getboolean('global', 'use_replication_slots'):
+            return
+        # We need to drop the slot in the old primary.
+        # But we don't know who the primary was (probably there are many of them).
+        # So, we need to release the lock on all hosts.
+        all_hosts = self.zk.get_children(self.zk.MEMBERS_PATH)
```
vicpopov, line 1150: Maybe loop over children of `self.zk.HOST_REPLICATION_SOURCES`?
It has only hosts that ever was a replications source (with locks), it should be faster than all hosts.

src/main.py
```diff
@@ -1121,6 +1171,7 @@ def _return_to_cluster(self, new_primary, role, is_dead=False):
             self.checks['primary_switch'] = 1
         logging.debug("primary_switch checks is %d", self.checks['primary_switch'])
 
+        self._acquire_replication_source_lock(new_primary)
```
vicpopov, line 1174: See comment on https://github.com/yandex/pgconsul/pull/30/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R657
It looks like a repeated call during single iteration.

tests/features/kill_primary.feature
```diff
@@ -344,7 +344,12 @@ Feature: Destroy primary in various scenarios
             state: streaming
         """
         When we stop container "postgresql3"
-        When we drop replication slot "pgconsul_postgresql3_1_pgconsul_pgconsul_net" in container "postgresql1"
+        And we wait "10.0" seconds
```
vicpopov, line 347: It would be a flapping test, since `drop_slot_countdown` is 10 too.
10 seconds is too soon to guarantee that 10 iterations have passed and slot was dropped.

tests/features/kill_primary.feature
vicpopov, line 1: Actually, the test `Destroy primary with one replica in archive recovery` now tests nothing.
You have implemented automatic restoring of slot, and now nothing prevents alive replica from streaming, and we can't be sure it is in archive mode.

vicpopov (changes requested):

tests/features/kill_primary.feature
munakoiso: There is retries in ```Then container "postgresql1" has following replication slots``` check

munakoiso: There are retries in ```Then container "..." has following replication slots``` check

src/main.py
```diff
@@ -646,8 +652,9 @@ def non_ha_replica_iter(self, db_state, zk_state):
                     wal_receiver_info and wal_receiver_info[0].get('status') == 'streaming'
                 )
                 logging.error(replication_source_replica_info)
-                current_primary = zk_state['lock_holder']
+
                 if replication_source_is_dead:
+                    self._acquire_replication_source_lock(current_primary)
```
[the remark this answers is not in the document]
munakoiso: It's an extra, you are right. Because there is _acquire_replication_source_lock call in _return_to_cluster function (below this line).
But we should acquire lock:
- When we are streaming, because if not, the lock will not be held, but the slot still exists, so if we don't acquire the lock, the replication slot will be dropped by the source.
- When we aren't streaming. In the event of long replica unavailability or a changing replication source.

src/main.py
```diff
@@ -776,6 +784,7 @@ def replica_iter(self, db_state, zk_state):
             if holder != db_state['primary_fqdn'] and holder != my_hostname:
                 self._replication_manager.leave_sync_group()
                 return self.change_primary(db_state, holder)
+            self._acquire_replication_source_lock(holder)
```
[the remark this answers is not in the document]
munakoiso: I think everything is right now.
https://github.com/yandex/pgconsul/pull/30#discussion_r1696408494

src/main.py
munakoiso: https://github.com/yandex/pgconsul/pull/30#discussion_r1696408494

tests/features/kill_primary.feature
munakoiso: I have made some changes to the test. I have disabled the 'use_replication_slots' option for host3, so a slot will not be created until we set the option back.

src/main.py
```diff
@@ -634,7 +637,13 @@ def non_ha_replica_iter(self, db_state, zk_state):
                 db_state['wal_receiver'],
                 replics_info,
             )
-            if not streaming and not can_delayed:
+            current_primary = zk_state['lock_holder']
+
+            if streaming_from_primary:
+                self._acquire_replication_source_slot_lock(current_primary)
+            if streaming:
```
vicpopov, line 644: `elif`? Or some other checks structure, because `current_primary` can be same as `stream_from`, and both `streaming_from_primary` and `streaming` can be true.

vicpopov (changes requested):

tests/features/kill_primary.feature
munakoiso: I think it makes the test more explicit. So, I want to leave it as it is.

src/main.py
```diff
@@ -1093,7 +1093,7 @@ def _attach_to_primary(self, new_primary, limit):
         return True
 
     def _handle_slots(self):
-        if not self.config.getboolean('global', 'use_replication_slots'):
+        if not self.config.getboolean('global', 'replication_slots_polling'):
```
vicpopov, line 1096: Although you have assigned config value of `replication_slots_polling` as value of `use_replication_slots`, it is not guaranteed that `replication_slots_polling` is True together with `replication_slots_polling`.
I think we should check both options, or forcefully set `replication_slots_polling` to False if `replication_slots_polling` is False.

vicpopov (changes requested):

vicpopov (dismissed):

vicpopov (approved):

PR #32, opened 2024-08-29
Remove redundant comment when handling replication slots

munakoiso:


reshke (dismissed):

reshke:
Maybe lower error msg to debug?

munakoiso:
> Maybe lower error msg to debug?

We call this function every iteration... I think even 'debug' is too much

PR #34, opened 2024-09-19
Checkpoint on primary before switchover

munakoiso:


src/main.py
```diff
@@ -1815,6 +1815,7 @@ def _do_primary_switchover(self, zk_state):
 
         # Attempt to shut down local PG instance.
         # Failure is not critical.
+        self.db.checkpoint()
```
vicpopov, line 1818: It won't help to reduce downtime for users of database this way, since pooler is stopped.
You should perform checkpoint before stopping pooler here: https://github.com/yandex/pgconsul/pull/34/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R1787

vicpopov (approved):

reshke (approved):

PR #35, opened 2024-09-20
MDB-30823: use replication lag in seconds instead of bytes, allow lag…

efimkin:
in scheduled switchover, replication lag is allowed, because primary will be stopped before replica has promote. recognize, agree?

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1843,16 +1843,18 @@ class pgconsul(object):
             return True
 
         replics_info = db_state.get('replics_info', list())
+        max_allowed_lag_ms = self.config.getint('global', 'max_allowed_switchover_lag_ms')
         for replica in replics_info:
             if replica.get('sync_state', '') != 'quorum':
                 continue
             if replica.get('application_name', '') != helpers.app_name_from_fqdn(switchover_candidate):
                 continue
-            replay_lag = replica.get('replay_location_diff', -1)
-            if replay_lag != 0:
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -1848,11 +1848,12 @@ def _candidate_is_sync_with_primary(self, db_state, switchover_candidate):
                 continue
             if replica.get('application_name', '') != helpers.app_name_from_fqdn(switchover_candidate):
                 continue
-            replay_lag = replica.get('replay_location_diff', -1)
-            if replay_lag != 0:
+            replay_lag = replica.get('replay_lag', -1)
```
vicpopov, line 1851: `replica`, which is a single record from `get_replics_info()` call, does not have `replay_lag` field, it is spelled as `replay_lag_msec`: https://github.com/yandex/pgconsul/blob/main/src/pg.py#L327

src/main.py
```diff
@@ -1848,11 +1848,12 @@ def _candidate_is_sync_with_primary(self, db_state, switchover_candidate):
...
+            logging.info(f"Replica {switchover_candidate} replay lag is {replay_lag} seconds"
+            if replay_lag > 10:
```
vicpopov, line 1853: I would suggest putting this `10` constant into config, so that we can override it there, instead of changing code.
Different clusters might require different guarantees of switchover time.

vicpopov (approved):

PR #39, opened 2024-10-09
release and recreate replication source lock when lock.acquire is failed

munakoiso:


reshke (approved):

src/zk.py
```diff
@@ -220,14 +220,22 @@ def _acquire_lock(self, name, allow_queue, timeout, read_lock=False):
...
-            return False
+            acquired = False
         except Exception:
             for line in traceback.format_exc().split('\n'):
                 logging.error(line.rstrip())
-            return False
+            acquired = False
+        if not acquired and release_on_fail:
+            logging.debug('Try to release and delete lock "%s", to recreate on next iter', name)
+            self.release_lock(name)
```
vicpopov, line 235: Actually, `release_lock()` can throw exceptions in case when there is some connection problem to ZK, and we are not catching it here.
IDK where it would be caught, but it seems like a wrong behaviour.

PR #47, opened 2024-11-06
MDB-32021: fix "Switchover is impossible now"

munakoiso:


Denchick (approved):

src/utils.py
```diff
@@ -113,14 +117,21 @@ def perform(self, min_replicas, block=True, timeout=None):
...
         return result
 
-    def in_progress(self, primary=None, timeline=None):
+    def in_progress(self, primary=None, timeline=None, return_true_on_zk_fail=False):
         """
         Return True if the cluster is currently in the process of switching
         over.
         Optionally check for specific hostname being currently the primary
         and having a particular timeline.
         """
```
vicpopov, line 126: Please add description of new parameter `return_true_on_zk_fail` to docstring.

src/main.py
```diff
@@ -1887,8 +1887,9 @@ def _transition_primary_switchover(self):
                 self.zk.delete('%s/%s/op' % (self.zk.MEMBERS_PATH, helpers.get_hostname()))
                 self._attach_to_primary(primary, self.config.getfloat('replica', 'recovery_timeout'))
                 return True
-        # Mark switchover node as failure
-        self.zk.write(self.zk.SWITCHOVER_STATE_PATH, 'replica_timed_out', need_lock=False)
+            logging.warning(f'SWITCHOVER_STATE_PATH ({self.zk.SWITCHOVER_STATE_PATH}) became None, but there is no one, who holds the leader lock.')
+            return False
+        logging.warning(f'SWITCHOVER_STATE_PATH ({self.zk.SWITCHOVER_STATE_PATH}) has value {self.zk.get(self.zk.SWITCHOVER_STATE_PATH)}, but expected to be None in timeout. Hope that the new master is doing well.')
```
vicpopov, line 1892: Actually, if we return False from `_transition_primary_switchover`, we would get into situation, when we are primary, not holding lock, and iteration is about to end.
So, on next iteration we would try to acquire lock and open old primary, thus failing switchover completely; there can be no luck for new primary.

This is not really a bad thing to indicate that with `replica_timed_out` state (similar to `master_timed_out`, which is not used too btw).

Maybe, it is better to solve current problem with "Switchover is impossible now" on the other end: in cli when checking   switchover state?
That would be the same: we just ignore this timed out outcome and try to perform switchover again.

vicpopov (changes requested):

src/main.py
munakoiso: The problem with replica_timed_out is possible race: _drop_stale_switchover on next iteration of new master deletes switchover.primary node, then old master marks switchover as replica_timed_out. So, after all, new switchover is impossible.

munakoiso: I removed master_timed_out too for more consistency. (but i can't imagine any possible problems/races with this state)

src/utils.py
```diff
@@ -135,15 +146,19 @@ def in_progress(self, primary=None, timeline=None):
             return state['progress']
         return False
 
-    def state(self):
+    def state(self, raise_zk_exceptions=False):
         """
         Current cluster state.
+        if raise_zk_exceptions is true - function will not except ZookeeperException
```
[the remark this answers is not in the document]
munakoiso: I mean "function will not catch ZookeeperException"
Fixed

src/main.py
vicpopov: > The problem with replica_timed_out is possible race: _drop_stale_switchover on next iteration of new master deletes switchover.primary node, then old master marks switchover as replica_timed_out. So, after all, new switchover is impossible.

Ok, now I finally understand. why this is a problem.
Why don't we fix that race? It is already half-fixed: we don't perform cleanup if we don't have lock: https://github.com/yandex/pgconsul/blob/main/src/main.py#L914
We just need to add lock acquiring before writing to `self.zk.SWITCHOVER_STATE_PATH`, and if this node is missing, just give up.
But since it is not used, we can just avoid writing, like you did here.

vicpopov (approved):

PR #48, opened 2024-11-06
MDB-32028: acquire_replication_source_slot_lock in all awaits

munakoiso:


Denchick (approved):

src/main.py
```diff
@@ -1528,13 +1528,15 @@ def _do_failover(self):
...
 
-    def _wait_for_recovery(self, limit=-1):
+    def _wait_for_recovery(self, new_primary, limit=-1):
         """
         Stop until postgresql complete recovery.
         With limit=-1 the loop here can be infinite.
         """
 
         def check_recovery_completion():
+            self.re_init_zk()
```
vicpopov, line 1538: I would suggest moving this call into `_acquire_replication_source_slot_lock` after check that `replication_slots_polling` feature is enabled.

PR #49, opened 2024-11-06
fix race in coordinator test

munakoiso:
both replicas might be async, so failover might be not possible

### DIFF
--- a/tests/features/coordinator.feature
+++ b/tests/features/coordinator.feature
@@ -47,6 +47,7 @@ Feature: Interacting with coordinator
           - client_hostname: pgconsul_postgresql3_1.pgconsul_pgconsul_net
             state: streaming
         """
+        Then container "postgresql1" replication state is "sync"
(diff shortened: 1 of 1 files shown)


tests/features/coordinator.feature
```diff
@@ -47,6 +47,7 @@ Feature: Interacting with coordinator
           - client_hostname: pgconsul_postgresql3_1.pgconsul_pgconsul_net
             state: streaming
         """
+        Then container "postgresql2" replication state is "sync"
```
vicpopov, line 50: `postgresql2` is not primary in this scenario, but `replication state is` check is for primary, it checks primary synchronous_standby_names: https://github.com/yandex/pgconsul/blob/main/tests/steps/database.py#L91
You should check `postgresql1` container here.

vicpopov (changes requested):

vicpopov (approved):

PR #51, opened 2024-11-08
enable maintenance in reset-all

munakoiso:


src/cli.py
```diff
@@ -92,16 +92,20 @@ def _wait_maintenance_disabled(zk, timeout):
     logging.info('Success')
 
 
+def enable_maintenance(zk, timeout, wait_all):
+    zk.ensure_path(zk.MAINTENANCE_PATH)
+    zk.noexcept_write(zk.MAINTENANCE_PATH, 'enable', need_lock=False)
+    if wait_all:
```
vicpopov, line 98: NIT: This check suggests that `wait_all` is more important than `timeout`, so `wait_all` should come first, and timeout should have a default, thus can be omitted.

src/cli.py
```diff
@@ -189,7 +193,7 @@ def reset_all(opts, conf):
     if all_nodes is None:
         logging.error("Could not get nodes to reset")
         all_nodes = []
-    nodes_to_delete = [x for x in all_nodes if x != zk.MEMBERS_PATH]
+    nodes_to_delete = [x for x in all_nodes if x not in (zk.MEMBERS_PATH, zk.MAINTENANCE_PATH)] + [zk.MAINTENANCE_PATH]
```
vicpopov, line 196: Do you mean here, that `zk.MAINTENANCE_PATH` node should be deleted last because it is checked first in every iteration, and it is a barrier against races with working pgconsul?
It is true when maintenance is enabled, but not true if not; there can be other races. (UPD.: now I see this is the case; it would be more clear to add `zk.MAINTENANCE_PATH` in the end, after main deletion cycle).
Generally, we need to establish an order of importance of this keys, and use it when cleaning (my #50 is actually about the same thing).

But that's for future discussion, not a topic for this PR.

vicpopov (approved):

PR #52, opened 2024-11-12
better logging in _handle_slots

munakoiso:


src/pg.py
```diff
@@ -736,24 +736,27 @@ def stop_postgresql(self, timeout=60):
                 logging.warning(line.rstrip())
         return self._cmd_manager.stop_postgresql(timeout, self.pgdata)
 
-    def replication_slots(self, action, slots):
-        """
-        Perform replication slots action (create/drop)
-        """
+    def get_replication_slots(self):
```
vicpopov, line 739: Why is this function needed?
It has no added value at all.
If you want to make `_get_replication_slots()` usable outside class, just rename it.

src/main.py
```diff
@@ -1174,16 +1174,31 @@ def _handle_slots(self):
...
         slot_names = [helpers.app_name_from_fqdn(fqdn) for fqdn in slot_lock_holders]
+        actual_replication_slots = self.db.get_replication_slots()
+        if actual_replication_slots is None:
+            logging.warning('Failed to get actual replication slots')
+            # However, we can continue here and try to create slots. None of slots will be dropped, but some might be created
+            actual_replication_slots = []
+        else:
+            logging.debug('Actual replication slots: %s', actual_replication_slots)
+
+        slot_names_to_create = [slot_name for slot_name in slot_names if slot_name not in actual_replication_slots]
```
vicpopov, line 1185: Why do we filter out already existing slots here?
It duplicates the same check inside `create_replication_slots`: https://github.com/yandex/pgconsul/pull/52/files#diff-1dd21d0846b6c985aaf231799772c4797a69429020c8d3781caf8e285d35db43R746
Maybe we need an option to quiet logging there?

vicpopov (changes requested):

vicpopov (approved):

vicpopov (approved):

PR #59, opened 2025-02-05
MDB-33721: extracted wal_reciever info, fix finish_iteration position and cascade replication

aak74:
Repairing pgconsul crash and fix delay when exception occurs

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -691,7 +691,7 @@ class pgconsul(object):
                 replication_source_streams = bool(
                     wal_receiver_info and wal_receiver_info.get('status') == 'streaming'
                 )
-                logging.error(replication_source_replica_info)
+                logging.error('replication_source_replica_info: {}'.format(replication_source_replica_info))
 
                 if replication_source_is_dead:
                     # Replication source is dead. We need to streaming from primary while it became alive and start streaming from primary.
@@ -1219,12 +1219,13 @@ class pgconsul(object):
         """
         Return to cluster (try stupid method, if it fails we try rewind)
         """
-        logging.info('ACTION. Starting returning to cluster. New primary: {}'.format(new_primary))
+        logging.info('ACTION. Starting return to cluster. New primary: {}'.format(new_primary))
         if self.checks['primary_switch'] >= 0:
             self.checks['primary_switch'] += 1
         else:
             self.checks['primary_switch'] = 1
 
+        logging.debug('primary switch checks is {}'.format(self.checks['primary_switch']))
         self._acquire_replication_source_slot_lock(new_primary)
         failover_state = self.zk.noexcept_get(self.zk.FAILOVER_INFO_PATH)
         if failover_state is not None and failover_state not in ('finished', 'promoting', 'checkpointing'):
@@ -1256,7 +1257,7 @@ class pgconsul(object):
             if role != 'primary' and not self.is_op_destructive(last_op) and not self._is_simple_primary_switch_tried():
                 logging.info('Trying to do a simple primary switch: {}'.format(new_primary))
                 result = self._try_simple_primary_switch_with_lock(limit, new_primary, is_dead)
-                logging.info('WAL source switch count: %s finish with result: %s', self.checks['primary_switch'], result)
+                logging.info('primary switch count: %s finish with result: %s', self.checks['primary_switch'], result)
                 if result:
                     self.checks['primary_switch'] = 0
                 return None
--- a/src/main.py
+++ b/src/main.py
@@ -1034,7 +1034,7 @@ class pgconsul(object):
         primary_switch_checks = self.config.getint('replica', 'primary_switch_checks')
         need_restart = self.config.getboolean('replica', 'primary_switch_restart')
 
-        logging.info('ACTION. Starting simple primary switch to {}'.format(new_primary))
+        logging.info('Starting simple primary switch to {}'.format(new_primary))
         if self.checks['primary_switch'] >= primary_switch_checks:
             self._set_simple_primary_switch_try()
 
@@ -1068,7 +1068,7 @@ class pgconsul(object):
                 #
                 # The easy way succeeded.
                 #
-                logging.info('ACTION. Simple switch primary to {} succeeded'.format(new_primary))
+                logging.info('Simple switch primary to {} succeeded'.format(new_primary))
                 self._reset_simple_primary_switch_try()
                 return True
             else:
--- a/src/main.py
+++ b/src/main.py
@@ -299,10 +299,12 @@ class pgconsul(object):
                 logging.error("Upper exception was for replica")
                 self.handle_detached_replica(db_state)
                 self.re_init_zk()
-                self.finish_iteration(timer)
             else:
                 self.re_init_zk()
+
+            self.finish_iteration(timer)
             return
+
         stream_from = self.config.get('global', 'stream_from')
         if role is None:
             self.dead_iter(db_state, zk_state, is_in_terminal_state=terminal_state)
--- a/src/main.py
+++ b/src/main.py
@@ -34,7 +34,7 @@ class pgconsul(object):
     DESTRUCTIVE_OPERATIONS = ['rewind']
 
     def __init__(self, **kwargs):
-        logging.debug('Initializing main class.')
+        logging.info('Initializing main class.')
         self.config = kwargs.get('config')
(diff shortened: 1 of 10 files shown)


fedusia (approved):

fedusia (approved):

src/main.py
```diff
@@ -636,6 +623,7 @@ def replica_return(self, db_state, zk_state):
         my_hostname = helpers.get_hostname()
         self.write_host_stat(my_hostname, db_state)
         holder = zk_state['lock_holder']
+        logging.debug('Replica was return. So we resume WAL replay to {}'.format(holder))
```
vicpopov, line 626: `Replica is returning`?
And it looks like logging an intent, shouldn't it be marked with `ACTION` and placed right before that action (https://github.com/yandex/pgconsul/pull/59/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R632)?

src/main.py
```diff
@@ -647,7 +635,7 @@ def replica_return(self, db_state, zk_state):
             # Wal receiver is not running and
             # postgresql isn't in archive recovery
             # We should try to restart
-            logging.warning('We should try switch primary one more time here.')
+            logging.warning('We should try switch source of WAL to {}'.format(holder))
```
vicpopov, line 638: `source of WAL` is a non-existing term.
Should we really introduce it?
Why not `We should try switch primary to {} again`?

src/main.py
```diff
@@ -1264,9 +1253,11 @@ def _return_to_cluster(self, new_primary, role, is_dead=False):
             last_op = self.zk.noexcept_get('%s/%s/op' % (self.zk.MEMBERS_PATH, helpers.get_hostname()))
             logging.info('Last op is: %s' % str(last_op))
             if role != 'primary' and not self.is_op_destructive(last_op) and not self._is_simple_primary_switch_tried():
-                logging.info('Trying to do a simple primary switch.')
+                logging.info('Trying to do a simple WAL source switch to: {}'.format(new_primary))
                 result = self._try_simple_primary_switch_with_lock(limit, new_primary, is_dead)
-                logging.info('Primary switch count: %s finish with result: %s', self.checks['primary_switch'], result)
+                logging.info('WAL source switch count: %s finish with result: %s', self.checks['primary_switch'], result)
```
vicpopov, line 1258: I think, naming a variable `primary_switch`, and logging it as `WAL source switch` will make debugging more obscure.

src/main.py
```diff
@@ -1264,9 +1253,11 @@ def _return_to_cluster(self, new_primary, role, is_dead=False):
...
+                if result:
+                    self.checks['primary_switch'] = 0
```
vicpopov, line 1260: There is already setting `primary_switch` to 0 inside `_try_simple_primary_switch_with_lock` before returning `True` here: https://github.com/yandex/pgconsul/pull/59/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R1048 and here: https://github.com/yandex/pgconsul/pull/59/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R1053

I think, we can either add it to the last possible place when True can be returned: https://github.com/yandex/pgconsul/pull/59/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R1077
Or we can remove it from everywhere and set it outside `_try_simple_primary_switch_with_lock` call.

Either way is good, but please not a mix of them.

src/replication_manager.py
```diff
@@ -68,6 +68,9 @@ def update_replication_type(self, db_state, ha_replics):
         needed = _get_needed_replication_type(self._config, self._db, db_state, ha_replics)
         logging.info('Needed replication type is %s.', needed)
 
+        if needed != current[0]:
+            logging.info('ACTION. We should change replication from {} to {}'.format(current[0], needed))
```
vicpopov, line 72: I think it is not an `ACTION`, since it is not immediately preceding the described switch.

src/replication_manager.py
```diff
@@ -90,10 +93,10 @@ def update_replication_type(self, db_state, ha_replics):
         else:
             logging.info("Here we should turn synchronous replication on.")
             if self._db.change_replication_to_sync_host(holder_fqdn):
-                logging.info('Turned synchronous replication ON.')
+                logging.info('ACTION. Turned synchronous replication ON.')
```
vicpopov, line 96: `ACTION` should be in https://github.com/yandex/pgconsul/pull/59/files#diff-85b283c50c288107310ead558a64a812605cafbb275800211c01540f2c8fd98fR94 log message, before performing action, not after.

src/main.py
```diff
@@ -636,18 +623,20 @@ def replica_return(self, db_state, zk_state):
         my_hostname = helpers.get_hostname()
         self.write_host_stat(my_hostname, db_state)
         holder = zk_state['lock_holder']
+        logging.debug('Replica is returning. So we resume WAL replay to {}'.format(holder))
```
vicpopov, line 626: This is superfluous, we have almost same message at [R632](https://github.com/yandex/pgconsul/pull/59/files#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R632), and it is better there as intent logging.

src/main.py
```diff
@@ -1084,13 +1074,15 @@ def _simple_primary_switch(self, limit, new_primary, is_dead):
...
+                logging.info('ACTION. Simple switch primary to {} succeeded'.format(new_primary))
+                self.checks['primary_switch'] = 0
+                self._reset_simple_primary_switch_try()
                 return True
             else:
                 return False
 
     def _rewind_from_source(self, is_postgresql_dead, limit, new_primary):
-        logging.info("Starting pg_rewind")
+        logging.info("ACTION. Starting pg_rewind")
```
vicpopov, line 1085: I think it is not an `ACTION`.
Here we are just beginning to process a rewind branch, with a lot of checks.
`Starting pg_rewind` intent should be logged right before calling `pg_rewind` executable.

src/main.py
```diff
@@ -1228,13 +1220,13 @@ def _return_to_cluster(self, new_primary, role, is_dead=False):
         """
         Return to cluster (try stupid method, if it fails we try rewind)
         """
-        logging.info('Starting returning to cluster.')
+        logging.info('ACTION. Starting return to cluster. New primary: {}'.format(new_primary))
```
vicpopov, line 1223: I think it is not an `ACTION`.
ACTION is a simple atomic operation, like switching some setting, service, running command, but not this big complex deed.

src/main.py
```diff
@@ -1262,11 +1254,17 @@ def _return_to_cluster(self, new_primary, role, is_dead=False):
             # rewinding and failed. So only hard way possible in this case.
             #
             last_op = self.zk.noexcept_get('%s/%s/op' % (self.zk.MEMBERS_PATH, helpers.get_hostname()))
-            logging.info('Last op is: %s' % str(last_op))
-            if role != 'primary' and not self.is_op_destructive(last_op) and not self._is_simple_primary_switch_tried():
-                logging.info('Trying to do a simple primary switch.')
+            if role == 'primary' or self.is_op_destructive(last_op) or self._is_simple_primary_switch_tried():
+                logging.info('Could not do a simple primary switch')           
+                logging.debug('Possible reasons: Role: %s, Last op is destructive: %s, Simple primary switch tried: %s',
+                    role, self.is_op_destructive(last_op), self._is_simple_primary_switch_tried()
```
vicpopov, line 1260: `self._is_simple_primary_switch_tried()` call is fetching data from ZK.
Calling it for logging only:
- slows down everything
- gives unpredictably different result from one that was used in `if` statement

src/pg.py
```diff
@@ -512,7 +514,7 @@ def _wait_for_primary_role(self):
         sleep_time = self.config.getfloat('global', 'iteration_timeout')
         role = self.get_role()
         while role != 'primary':
-            logging.debug('Our role should be primary but we are now "%s".', role)
+            logging.info('ACTION. Our role should be primary but we are now "%s".', role)
```
vicpopov, line 517: How is this an `ACTION`? What are we going to do after this log record?

vicpopov (changes requested):

vicpopov (approved):

PR #65, opened 2025-04-21
MDB-35514: added mypy, removed implicit dependencies from several classes

aak74:


src/command_manager.py
```diff
@@ -11,16 +11,16 @@
...
 class CommandManager:
-    def __init__(self, config):
-        self._config = config
+    def __init__(self, commands: dict[str, str]):
+        self._commands = commands
 
-    def _prepare_command(self, command_name, **kwargs):
-        command = self._config.get('commands', command_name)
+    def _prepare_command(self, command_name: str, **kwargs):
+        command = self._commands.get(command_name, '') # type: str
```
vovik0134, line 18: You can define type explicitly without using hints to mypy. Like this:
```python
command: str = self._commands.get(command_name, '')
```

src/__init__.py
```diff
@@ -7,10 +7,10 @@
 import os
 import sys
 
-try:
-    from configparser import RawConfigParser
-except ImportError:
+if sys.version_info[0] < 3:
```
vovik0134, line 10: This predicate looks obsolete, cause there are a lot of examples of type hints usage that became available in python 3.5. Hence, code (as a whole) won't work under python2, therefore there is no need to use this predicate

src/failover_election.py
```diff
@@ -48,10 +51,10 @@ class FailoverElection(object):
     def __init__(
         self,
         config,
```
vovik0134, line 53: Is it possible to add type hints for other parameters here as well?

src/main.py
```diff
@@ -53,25 +58,57 @@ def __init__(self, **kwargs):
...
+                self._replication_manager_config(),
                 self.db,
                 self.zk,
             )
 
     def _sigterm_handler(self, *_):
         self._should_run = False
 
+    def _commands(self) -> dict[str, str]:
+        return dict(self.config.items('commands')) if self.config.has_section('commands') else {}
```
vovik0134, line 83: Using inline `if ... else` statement (ternary operator) considered as a bad practice. It's better to write it as simple if statement. Like this:
```python
def _commands(self) -> dict[str, str]:
    if self.config.has_section('commands'):
        return self.config

    return {}
```

vovik0134, line 83: There is no need to construct dict from existing dict (it's already dict). If you intended to make a copy of an existing dict, then it's better to use `deepcopy` function

src/pg.py
```diff
@@ -172,14 +186,14 @@ def reconnect(self):
...
-            self.conn_local = psycopg2.connect(self.config.get('global', 'local_conn_string'))
+            self.conn_local = psycopg2.connect(self.config.conn_string)
             self.conn_local.autocommit = True
 
             self.role = self.get_role()
             self.pg_version = self._get_pg_version()
             self.pgdata = self._get_pgdata_path()
         except psycopg2.OperationalError:
-            logging.error('Could not connect to "%s".', self.config.get('global', 'local_conn_string'))
+            logging.error('Could not connect to "%s".', self.config.conn_string)
```
vovik0134, line 196: Not sure if it's good idea to place `conn_string` as is, because it may content sensitive data (e.g. password)

src/main.py
aak74: - `self.config` is dict.
- `self.config.items('commands')` is list[tuple[str, str]]

I need to use `self.config.items('commands')` so `dict(self.config.items('commands'))`

src/pg.py
aak74: May be later.
Here just small refactoring.

Makefile
```diff
@@ -110,3 +110,9 @@ check: build check_test
 check_unstoppable: build check_test_unstoppable
 
 check-world: clean build check_test jepsen_test
+
+mypy-install:
```
vovik0134, line 114: Not sure if this is correct to install `mypy` separately, as it's already defined in `requirements.txt` (with pinned version, btw), and in `Makefile` we already have section with `requirements.txt` usage

src/failover_election.py
```diff
@@ -82,7 +83,9 @@ def _collect_votes(self):
         votes = {}
         app_name_map = {helpers.app_name_from_fqdn(host): host for host in self._zk.get_ha_hosts()}
         for info in self._replica_infos:
-            app_name = info['application_name']
+            app_name = info.get('application_name')
```
vovik0134, line 86: Here behaviour has changed. In previous version we got `KeyError`, when info has no `application_name`  key. Now, with `get` `dict`-method we don't get any error here.

Is this change in behaviour intended? If so, what issue does it solve?

src/main.py
```diff
@@ -53,25 +55,60 @@ def __init__(self, **kwargs):
...
+    def _get_repllication_manager(self) -> ReplicationManager:
         if self.config.getboolean('global', 'quorum_commit'):
-            self._replication_manager = QuorumReplicationManager(
-                self.config,
+            return QuorumReplicationManager(
+                self._replication_manager_config(),
                 self.db,
                 self.zk,
             )
         else:
```
vovik0134, line 69: nit: Like in `go` you can avoid `else` blocks, if `return` is present in `then`-block. Like this:
```python
def _get_repllication_manager(self) -> ReplicationManager:
    if self.config.getboolean('global', 'quorum_commit'):
        return QuorumReplicationManager(self._replication_manager_config(), self.db, self.zk)
    
    return SingleSyncReplicationManager(self._replication_manager_config(), self.db, self.zk)
```

This makes code simplier to read

vovik0134 (commented):
Second iteration of review

.github/workflows/docker-mypy.yml
```diff
@@ -0,0 +1,24 @@
+name: Mypy tests
+
+on:
+  workflow_dispatch:
+  schedule:
```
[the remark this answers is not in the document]
aak74: It work as expected.
On each PR.

BTW extract all flow `On PR` into specific config.

src/failover_election.py
aak74: You are right. There is no need for changes.

Makefile
```diff
@@ -110,3 +110,9 @@ check: build check_test
...
+	pip install mypy types-PyYAML
+
+mypy: mypy-install
```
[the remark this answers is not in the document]
aak74: You are right, but those linters not runs for now...
May be later we can make them works.

vovik0134 (approved):

src/replication_manager.py
```diff
@@ -54,9 +143,7 @@ def should_close(self):
                     should_wait = True
                     logging.debug('Replica %s has reply_time less than _zk_fail_timestamp %d', replica['client_hostname'], self._zk_fail_timestamp)
             if should_wait:
-                primary_unavailability_timeout = self._config.getfloat('replica', 'primary_unavailability_timeout')
-                logging.debug('We should wait primary_unavailability_timeout %d and check once more.', primary_unavailability_timeout)
```
munakoiso, line 58: we should return debug here

src/pg.py
```diff
@@ -676,7 +691,7 @@ def stop_archiving_wal_stopped(self):
 
     def _get_postgresql_auto_conf(self):
         config = {}
-        current_file = os.path.join(self.pgdata, 'postgresql.auto.conf')
+        current_file = os.path.join(self.pgdata or '', 'postgresql.auto.conf')
```
munakoiso, line 694: We can check of pgdata is empty when reading config, just once

src/zk.py
```diff
@@ -250,8 +250,11 @@ def _get_lock(self, name, read_lock) -> Lock:
     def _delete_lock(self, name: str) -> bool:
         if name in self._locks:
             del self._locks[name]
+            return True
```
munakoiso, line 253: mb it will be better to return nothing in this function? there is just one call of _delete_lock, and return is not checked

src/main.py
```diff
@@ -1816,7 +1854,7 @@ def _check_primary_switchover(self, db_state, zk_state):
             logging.warning(
                 'Last role transition was %.1f seconds ago,'
                 ' and alive host count less than HA hosts in zk (HA: %d, ZK: %d) ignoring switchover.',
-                time.time() - last_role_transition_ts,
+                time.time() - (last_role_transition_ts or 0),
```
munakoiso, line 1857: why this change is necessary?

src/plugin.py
```diff
@@ -93,7 +93,7 @@ def load_plugins(path):
     if path not in sys.path:
         sys.path.insert(0, path)
 
-    ret = {'Postgres': [], 'Zookeeper': []}
+    ret: dict[str, list] = {'Postgres': [], 'Zookeeper': []}
```
vicpopov, line 96: What is the type of element in these lists?

src/replication_manager.py
```diff
@@ -142,7 +229,7 @@ def change_replication_to_async(self):
             return True
         return False
 
-    def enter_sync_group(self, replica_infos):
+    def enter_sync_group(self, replica_infos: list[dict]):
```
vicpopov, line 232: what are the types inside this dict?

src/sdnotify.py
```diff
@@ -26,14 +32,15 @@ def __init__(self, debug=False):
...
         """Send string `msg` as bytes on the notification socket"""
-        if self.enabled():
-            try:
-                self.socket.sendall(msg.encode())
-            except Exception:
-                if self.debug:
-                    raise
-
-    def enabled(self):
+        if self.socket is None:
```
vicpopov, line 35: Should either use `enabled()` method here, or drop otherwise unused `enabled()`

aak74: fixed

src/zk.py
aak74: fixed

src/failover_election.py
```diff
@@ -47,17 +50,15 @@ class FailoverElection(object):
...
-        timeout,
-        replics_info,
-        replication_manager,
-        allow_data_loss,
-        host_priority,
-        host_lsn,
-        quorum_size,
+        _zk: Zookeeper,
+        timeout: int,
+        replics_info: list[dict],
```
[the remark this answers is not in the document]
aak74: fixed

src/helpers.py
```diff
@@ -118,7 +121,7 @@ def get_oldest_replica(replics_info):
     return None
 
 
-def make_current_replics_quorum(replics_info, alive_hosts):
+def make_current_replics_quorum(replics_info: list[dict], alive_hosts):
```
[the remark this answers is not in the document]
aak74: fixed

src/main.py
```diff
@@ -53,25 +55,61 @@ def __init__(self, **kwargs):
...
+            timeout=self.config.getfloat('global', 'iteration_timeout'),
+            wals_count_to_upload=self.config.getint('plugins', 'wals_to_upload'),
+        )
+
+    def _replication_manager_config(self) -> ReplicationManagerConfig:
+        return ReplicationManagerConfig(
+            priority = self.config.getint('global', 'priority'),
+            primary_unavailability_timeout = self.config.getfloat('replica', 'primary_unavailability_timeout'),
+            change_replication_metric = self.config.get('primary', 'change_replication_metric'),
+            metric = self.config.get('primary', 'change_replication_metric'),
```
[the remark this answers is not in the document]
aak74: fixed

src/command_manager.py
```diff
@@ -11,16 +11,16 @@
...
 class CommandManager:
-    def __init__(self, config):
-        self._config = config
+    def __init__(self, commands: dict[str, str]):
+        self._commands = commands
 
-    def _prepare_command(self, command_name, **kwargs):
-        command = self._config.get('commands', command_name)
+    def _prepare_command(self, command_name: str, **kwargs):
+        command: str = self._commands.get(command_name, '')
```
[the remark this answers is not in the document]
aak74: foxed

src/types.py
```diff
@@ -0,0 +1 @@
+ReplicaInfos = list[dict[str, object]]
```
vicpopov, line 1: I think it would be better to introduce a singular type `ReplicaInfo`, and use `List[ReplicaInfo]` in annotations. Or just introduce 2 types.

vicpopov, line 1: Maybe instead of `object` enumerate all possible values? There are not so much of them: int, str and that's all.

src/plugin.py
```diff
@@ -93,7 +93,7 @@ def load_plugins(path):
     if path not in sys.path:
         sys.path.insert(0, path)
 
-    ret: dict[str, list] = {'Postgres': [], 'Zookeeper': []}
+    ret: dict[str, list[PostgresPlugin | ZookeeperPlugin]] = {'Postgres': [], 'Zookeeper': []}
```
vicpopov, line 96: I think it's wrong type: we can't mix PG and ZK plugins in one list. It should be:
`list[PostgresPlugin] | list[ZookeeperPlugin]`

src/plugin.py
```diff
@@ -31,7 +31,7 @@ def before_promote(self, conn, config):
         """
         pass
 
-    def after_promote(self, conn, config):
+    def after_promote(self, conn, wals_count_to_upload: int):
```
vicpopov, line 34: And that is wrong: generic PostgresPlugin is not limited to deal with `wals_count_to_upload` only. It can perform some other actions, and entire config is more useful for it here.

src/pg.py
```diff
@@ -429,7 +430,7 @@ def check_walsender(self, replics_info, holder_fqdn):
             try:
                 if replica['sync_state'] == 'sync' and replica['application_name'] != holder_app_name:
                     logging.warning('It seems sync replica and sync replica holder are different. Killing walsender.')
-                    os.kill(replica['pid'], signal.SIGTERM)
+                    os.kill(replica['pid'], signal.SIGTERM) # type: ignore
```
vicpopov, line 433: See comment on `types.py`: if hints are more elaborate, we wouldn't need to ignore typing checks here.

vicpopov (changes requested):

src/main.py
```diff
@@ -53,25 +56,80 @@ def __init__(self, **kwargs):
...
+        return SingleSyncReplicationManager(
+            self._replication_manager_config(),
+            self.db,
+            self.zk,
+        )
+
     def _sigterm_handler(self, *_):
         self._should_run = False
 
+    def _commands(self) -> Commands:
```
vicpopov, line 80: I think this method should go to `command_manager` module, maybe as `__init__` for `Commands` class.
And we can pass entire config along with the name of section to get values from it.

src/command_manager.py
```diff
@@ -11,18 +12,34 @@
 }
 
 
+@dataclass
+class Commands:
```
vicpopov, line 16: Maybe `CommandsConfig`, as it actually holds `commands` config section from global config, but in a class form.

src/main.py
```diff
@@ -53,25 +56,80 @@ def __init__(self, **kwargs):
...
+                pooler_stop=self.config.get('commands', 'pooler_stop'),
+                pooler_status=self.config.get('commands', 'pooler_status'),
+                list_clusters=self.config.get('commands', 'list_clusters'),
+                generate_recovery_conf=self.config.get('commands', 'generate_recovery_conf'),
+            )
+
+        raise ValueError('No commands section in config')      
+
+    def _postgres_config(self) -> PostgresConfig:
+        return PostgresConfig(
```
vicpopov, line 100: Maybe it is a `GlobalConfig`, since all options are from `global` section?
Or just a `Config`, if we add other configs into it?

src/main.py
```diff
@@ -53,25 +56,80 @@ def __init__(self, **kwargs):
...
+            use_lwaldump=self.config.getboolean('global', 'use_lwaldump') or self.config.getboolean('global', 'quorum_commit'),
+            working_dir=self.config.get('global', 'working_dir'),
+            recovery_filepath=self.config.get('global', 'recovery_conf_rel_path'),
+            use_replication_slots=self.config.getboolean('global', 'use_replication_slots'),
+            standalone_pooler=self.config.getboolean('global', 'standalone_pooler'),
+            pooler_addr=self.config.get('global', 'pooler_addr'),
+            pooler_port=self.config.getint('global', 'pooler_port'),
+            pooler_conn_timeout=self.config.getfloat('global', 'pooler_conn_timeout'),       
+            postgres_timeout=self.config.getfloat('global', 'postgres_timeout'),
+            timeout=self.config.getfloat('global', 'iteration_timeout'),
```
vicpopov, line 111: It's a bad rename, we lose understanding which timeout out of like 10 it is.
Please, keep it as `iteration_timeout`, it is a well-known term in this codebase.

vicpopov (changes requested):

src/main.py
aak74: I don`t think so.
CommandManager don`t need to know about whole config.
May be we need to move this function into `__init__.py`.
But may be later.

src/command_manager.py
aak74: [Here](https://github.com/yandex/pgconsul/pull/65/files#diff-38e28b0ed01217afe2c8bff7eb3a35b6a20f9b38d21297c063068a4c5b7186b3R33) we definitely expects commands, not any config.

src/main.py
aak74: It`s config for Postgres. And nothing more.

src/main.py
aak74: Fixed

vicpopov (approved):
Good enough to merge.

PR #68, opened 2025-04-28
CLOUDDUTY-37357: fix bug during gettting wal receiver status

fedusia:


src/main.py
```diff
@@ -686,7 +686,7 @@ def non_ha_replica_iter(self, db_state, zk_state):
                 wal_receiver_info = self._zk_get_wal_receiver_info(stream_from)
                 logging.debug('wal_receiver_info: {}'.format(wal_receiver_info))
                 replication_source_streams = bool(
-                    wal_receiver_info and wal_receiver_info.get('status') == 'streaming'
+                    wal_receiver_info and wal_receiver_info[0].get('status') == 'streaming'
```
vicpopov, line 689: After #59 is is not a list, bug a plain dict: https://github.com/yandex/pgconsul/pull/59/files#diff-1dd21d0846b6c985aaf231799772c4797a69429020c8d3781caf8e285d35db43R368.
No back transformation to the list is performed anywhere.

It looks like a transition-time error, we must accept reading old-format value, but write only new format.
Maybe it's forever :-(

vicpopov (changes requested):

PR #69, opened 2025-04-29
MDB-33912 fix race between switchover cli and drop_stale_switchover function

munakoiso:


vicpopov (changes requested):

src/zk.py
```diff
@@ -225,18 +229,18 @@ def _acquire_lock(self, name, allow_queue, timeout, read_lock=False):
...
             try:
-                self.release_lock(name)
+                self.release_lock(name, host_description=host_description)
             except Exception:
                 for line in traceback.format_exc().split('\n'):
                     logging.error(line.rstrip())
         return acquired
 
-    def _get_lock(self, name, read_lock) -> Lock:
+    def _get_lock(self, name, read_lock, host_description=None) -> Lock:#
```
vicpopov, line 238: `#` in the end of line, something strange...

src/zk.py
```diff
@@ -186,26 +186,30 @@ def _write(self, path, data, need_lock=True):
...
+    def _init_lock(self, name, read_lock=False, host_description=None):
+        if host_description is None:
+            host_description = helpers.get_hostname()
         path = self._path_prefix + name
         if read_lock:
-            lock = self._zk.ReadLock(path, helpers.get_hostname())
+            lock = self._zk.ReadLock(path, host_description)
         else:
-            lock = self._zk.Lock(path, helpers.get_hostname())
+            lock = self._zk.Lock(path, host_description)
```
vicpopov, line 196: Sorry, but the same inconsistency as in https://github.com/yandex/pgconsul/pull/69#discussion_r2068343040 is still here.
We can init lock on some path with one `host_description`, and later try get it with `_get_lock` with another `host_description`, but will get lock instance with first `host_description`.

To fix this you need to rework `self._locks` map, making a tuple `(name, host_description)` key in it.

vicpopov (changes requested):

src/zk.py
munakoiso: Host description in one instance of pgconsul is constant, there is no such problem

src/zk.py
munakoiso: oops

src/zk.py
vicpopov: > Host description in one instance of pgconsul is constant, there is no such problem

Then it should be put it into `Zookeeper` class constructor, like I've suggested in https://github.com/yandex/pgconsul/pull/69#discussion_r2068343040

It's all about usability and predictability of this class. For example:
Given that a lock on path `path/to/lock` is held by `fqdn2`, if you call consecutively
```
zk.try_acquire_lock(path='path/to/lock', host_description='fqdn1')
zk.try_acquire_lock(path='path/to/lock', host_description='fqdn2')
```
then in second call the same lock object will be found in `_get_lock`, and it will work incorrectly with `host_description='fqdn1'`, and won't realize that lock is already held and whole call should return success instead of failure.

src/zk.py
```diff
@@ -509,14 +513,14 @@ def acquire_lock(self, lock_type, allow_queue=False, timeout=None, read_lock=Fal
...
+    def try_acquire_lock(self, lock_type=None, allow_queue=False, timeout=None, read_lock=False, host_description=None):
         """
         Acquire lock (leader by default)
         """
         lock_type = lock_type or self.PRIMARY_LOCK_PATH
-        return self._acquire_lock(lock_type, allow_queue, timeout, read_lock=read_lock)
+        return self._acquire_lock(lock_type, allow_queue, timeout, read_lock=read_lock, host_description=host_description)
 
-    def release_lock(self, lock_type=None, wait=0):
+    def release_lock(self, lock_type=None, wait=0, host_description=None):
```
vicpopov, line 523: Why do we need parameter `host_description` here, if we don't use it in this method?

src/zk.py
```diff
@@ -509,14 +513,14 @@ def acquire_lock(self, lock_type, allow_queue=False, timeout=None, read_lock=Fal
             raise ZookeeperException(f'Failed to acquire lock {lock_type}')
         logging.debug(f'Success acquire lock: {lock_type}')
 
-    def try_acquire_lock(self, lock_type=None, allow_queue=False, timeout=None, read_lock=False):
+    def try_acquire_lock(self, lock_type=None, allow_queue=False, timeout=None, read_lock=False, host_description=None):
```
vicpopov, line 516: Parameter `host_description` was not added to `def acquire_lock` method, and it is almost the same wrapper around `_acquire_lock`. Shouldn't we add it for the sake of symmetry?

src/zk.py
```diff
@@ -106,6 +101,19 @@ def __init__(self, config, plugins):
             for line in traceback.format_exc().split('\n'):
                 logging.error(line.rstrip())
 
+
+    def get_lock_contender_name(self):
```
vicpopov, line 105: We can calculate `self._lock_contender_name` during class initialization only once, and get rid of this method, using just `self._lock_contender_name` property when needed.

src/zk.py
```diff
@@ -546,28 +554,28 @@ def release_if_hold(self, lock_type, wait=0, read_lock=False):
...
+        return self._get_host_path(self.HOST_PRIO_PATH, hostname)
 
     def get_simple_primary_switch_try_path(self, hostname=None):
-        return _get_host_path(self.SIMPLE_PRIMARY_SWITCH_TRY_PATH, hostname)
+        return self._get_host_path(self.SIMPLE_PRIMARY_SWITCH_TRY_PATH, hostname)
 
     def get_election_vote_path(self, hostname=None):
         if hostname is None:
-            hostname = helpers.get_hostname()
+            hostname = self.get_lock_contender_name()
```
vicpopov, line 578: I'm not sure if this is the right place to move from plain hostname to lock_contender_name.

src/zk.py
```diff
@@ -106,6 +101,19 @@ def __init__(self, config, plugins):
...
+
+    def get_lock_contender_name(self):
+        if self._lock_contender_name:
+            return self._lock_contender_name
+        return helpers.get_hostname()
+
+
+    def _get_host_path(self, path, hostname):
+        if hostname is None:
+            hostname = self.get_lock_contender_name()
```
vicpopov, line 113: I think it is not what we need.
`_get_host_path` was meant to manipulate with `path` template and put there a hostname.
Nothing to do with locks and contender names here.

src/zk.py
munakoiso: hostname may be incorrect when starting the VM, so I prefer to call the hostname() function every time.

src/utils.py
```diff
@@ -8,10 +8,11 @@
 import logging
 import time
 from operator import itemgetter
+from os import getpid
 
 from . import read_config, zk
 from .exceptions import SwitchoverException, FailoverException
-from .helpers import app_name_from_fqdn
+from .helpers import app_name_from_fqdn, get_hostname
```
vicpopov, line 15: NIT: after importing `get_hostname`, it is used in both ways: `get_hostname()` and `helpers.get_hostname()`, not qute cute :(

vicpopov (approved):

src/utils.py
vicpopov: > NIT: after importing `get_hostname`, it is used in both ways: `get_hostname()` and `helpers.get_hostname()`, not qute cute :(

No, it's not one big file, it's different file, so it is irrelevant.

PR #71, opened 2025-05-05
MDB-35006 + do not change replication_type to async if zk is unavailable

munakoiso:
\+ better logging

### DIFF
--- a/src/replication_manager.py
+++ b/src/replication_manager.py
@@ -33,13 +33,17 @@ class SingleSyncReplicationManager:
             for replica in info:
                 if replica['reply_time_ms'] / 1000 < self._zk_fail_timestamp:
                     should_wait = True
+                    logging.debug('Replica %s has reply_time less than _zk_fail_timestamp %d', replica['client_hostname'], self._zk_fail_timestamp)
             if should_wait:
-                time.sleep(self._config.getfloat('replica', 'primary_unavailability_timeout'))
+                primary_unavailability_timeout = self._config.getfloat('replica', 'primary_unavailability_timeout')
+                logging.debug('We should wait primary_unavailability_timeout %d and check once more.', primary_unavailability_timeout)
+                time.sleep(primary_unavailability_timeout)
                 info = self._db.get_replics_info(self._db.role)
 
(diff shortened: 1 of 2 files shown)


tests/steps/cluster.py
```diff
@@ -849,6 +849,7 @@ def step_container_is_in_quorum_group(context, name):
         {
             fqdn: {
                 'state': 'streaming',
+                'sync_state': 'quorum',
```
vicpopov, line 852: Is it `'quorum'` for `'sync'` replication type?
Maybe we should use `name` here?

tests/features/autofailover.feature
```diff
@@ -122,6 +122,8 @@ Feature: Check pgconsul with disabled autofailover
 
         """
         Then <lock_type> "<lock_host>" has holder "pgconsul_postgresql1_1.pgconsul_pgconsul_net" for lock "/pgconsul/postgresql/leader"
+        And container "postgresql2" is in <replication_type> group
+        And container "postgresql3" is in <replication_type> group
         And <lock_type> "<lock_host>" has following values for key "/pgconsul/postgresql/replics_info"
```
vicpopov, line 127: Do we still need this check now?
It is a repetition of `container "..." is in ... group` actually, but without check for `sync_state` field.

src/replication_manager.py
```diff
@@ -335,6 +339,7 @@ def _get_needed_replication_type(config, db, db_state, ha_replics):
 
     if 'count' in metric:
         if replics_number == 0:
+            logging.debug("Needed repl type is async, because there is no alive replicas")
```
vicpopov, line 342: `alive` -> `streaming ha` to be exact in message.

src/replication_manager.py
```diff
@@ -354,6 +361,9 @@ def _get_needed_replication_type(config, db, db_state, ha_replics):
         except Exception:
             ratio = 0.0
         if ratio >= over:
+            logging.debug("Needed repl type is async, because current load ratio %f > overload_sessions_ratio %f",
```
vicpopov, line 364: `load ratio` -> `sessions ratio` to indicate exact meaning of metric (`load` term is too vague).

vicpopov (approved):

PR #74, opened 2025-05-07
MDB-35088: await before changing replication type to async

munakoiso:
In some cases hosts uses two different network interfaces: for zk connection and for replication. If there is 2-hosts (A,B) cluster and master (A) became unavailable not instantly (firstly - broken replication, then - zk connection): A will switch replication type to async instantly and write to zk that B is no longer sync/quorum replica. So, when zk connection also became broken and host A became fully unavailable - there might be some transactions that was commited into A, but not replicated to B. From this moment this transactions may be lost forever (if we can't fetch data from A), and host B can not become master.

So, there is why i added timeout before switching to async. In the situation described above A will wait _before_async_unavailability_timeout_ and maybe it will be enough to A to became fully unavailable (or available, who knows)

### DIFF
--- a/src/replication_manager.py
+++ b/src/replication_manager.py
@@ -4,8 +4,26 @@ import time
 
 from . import helpers
 
+class NeededReplicationTypeMixin:
+    def __init__(self):
+        self._async_waiting_timestamp = None
 
-class SingleSyncReplicationManager:
+
+    def _get_needed_replication_type(self, config, db, db_state, ha_replics):
+        replication_type = _get_needed_replication_type_without_await_before_async(config, db, db_state, ha_replics)
+        if replication_type == 'async':
+            now = time.time()
+            if self._async_waiting_timestamp is None:
+                self._async_waiting_timestamp = now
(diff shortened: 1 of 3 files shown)


src/replication_manager.py
```diff
@@ -322,7 +340,7 @@ def get_ensured_sync_replica(self, replica_infos):
         return sync_quorum.get(helpers.get_oldest_replica(quorum_info))
 
 
-def _get_needed_replication_type(config, db, db_state, ha_replics):
+def _get_needed_replication_type_without_await_before_async(config, db, db_state, ha_replics):
```
vicpopov, line 343: Ugh :(
But no better name comes to mind.
Maybe switch names with class method: leave here untouched and `_get_needed_replication_type_with_await_before_async` in mixin.

vicpopov (approved):

src/replication_manager.py
munakoiso: Seems like pretty same solution

vicpopov: > Seems like pretty same solution

Yep, not much better. But at least less negations.

PR #76, opened 2025-05-20
MDB-36033: for single node cluster try to start postgres

aak74:
- for single node cluster try to start postgres
- better logging fn get_control_parameter

### DIFF
--- a/src/command_manager.py
+++ b/src/command_manager.py
@@ -1,3 +1,5 @@
+import logging
+
 from . import helpers
 
 
@@ -35,7 +37,14 @@ class CommandManager:
         res = helpers.subprocess_popen(command, log_cmd=log)
         if not res:
             return None
-        value = res.communicate()[0].decode('utf-8').split(':')[-1].strip()
+        (stdout, stderr) = res.communicate()
+        if stderr:
+            logging.error('error occured with command %s', stderr.decode('utf-8'))
+            logging.error('stderr: %s', stderr.decode('utf-8'))
+            logging.error('stdout: %s', stdout.decode('utf-8'))
+            return None
+            
+        value = stdout.decode('utf-8').split(':')[-1].strip()
         if preproc:
             return preproc(value)
         else:
--- a/src/main.py
+++ b/src/main.py
@@ -848,12 +848,13 @@ class pgconsul(object):
             return None
 
         self.db.pgpooler('stop')
+        if self._is_single_node:
+            return self.db.start_postgresql()
 
         self._replication_manager.leave_sync_group()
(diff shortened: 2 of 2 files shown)


src/command_manager.py
```diff
@@ -35,7 +37,14 @@ def get_control_parameter(self, pgdata, parameter, preproc=None, log=True):
         res = helpers.subprocess_popen(command, log_cmd=log)
         if not res:
             return None
-        value = res.communicate()[0].decode('utf-8').split(':')[-1].strip()
+        (stdout, stderr) = res.communicate()
+        if stderr:
```
vicpopov, line 41: Something in stderr does not always indicate an error, it can be a warning.
I think, we should analyze `res.returncode` to detect error condition here.

vicpopov (changes requested):

src/command_manager.py
```diff
@@ -35,7 +37,14 @@ def get_control_parameter(self, pgdata, parameter, preproc=None, log=True):
...
+            logging.error('error occured with command %s', stderr.decode('utf-8'))
+            logging.error('stderr: %s', stderr.decode('utf-8'))
+            logging.error('stdout: %s', stdout.decode('utf-8'))
+            return None
+            
+        value = stdout.decode('utf-8').split(':')[-1].strip()
```
fedusia, line 47: maybe a better name for the variable timeline_id?

src/command_manager.py
```diff
@@ -35,7 +37,14 @@ def get_control_parameter(self, pgdata, parameter, preproc=None, log=True):
         res = helpers.subprocess_popen(command, log_cmd=log)
         if not res:
             return None
-        value = res.communicate()[0].decode('utf-8').split(':')[-1].strip()
+        (stdout, stderr) = res.communicate()
+        if stderr:
+            logging.error('error occured with command %s', stderr.decode('utf-8'))
```
fedusia, line 42: I think that here we want to log command name not stderr

vicpopov: > I think that here we want to log command name not stderr

Not necessary, actually, `helpers.subprocess_popen` logs command before launching it

src/main.py
```diff
@@ -848,12 +848,13 @@ def dead_iter(self, db_state, zk_state, is_in_terminal_state):
             return None
 
         self.db.pgpooler('stop')
+        if self._is_single_node:
```
vicpopov, line 851: I think we should log here why we did this decision to start postgres.

src/command_manager.py
aak74: > `helpers.subprocess_popen` logs command before launching it

Here we only have the log: `Called: get_control_parameter`

vicpopov (approved):

PR #79, opened 2025-06-06
MDB-36196: Fix behaviour with unexpected network inconsistency

aak74:
* Added test with unexpected network inconsistency
* Better logging
* Added rsync.log to extracted logs


src/main.py
```diff
@@ -1101,6 +1101,7 @@ def _simple_primary_switch(self, limit, new_primary, is_dead):
         logging.info('Starting simple primary switch to {}'.format(new_primary))
         if self.checks['primary_switch'] >= primary_switch_checks:
             self._set_simple_primary_switch_try()
+        self.db.enable_wal_receiver()
```
mialinx, line 1104: Looks like a possible data loss.
Just after failover and loosing election, this host still have old master in recovery.conf.
`enable_wal_receiver` will make the replica to continue streaming from old master.
It may confirm more LSNs, and they will be lost.

We can enable wal receiver only after updating recovery.conf

src/main.py
```diff
@@ -681,10 +685,6 @@ def replica_return(self, db_state, zk_state):
         self.checks['failover'] = 0
         limit = self.config.getfloat('replica', 'recovery_timeout')
 
-        # Try to resume WAL replaying, it can be paused earlier
-        logging.debug('ACTION. Replica is returning. So we resume WAL replay to {}'.format(holder))
-        self.db.pg_wal_replay_resume()
-
```
aak74, line 687: We don`t want run it here.
It will be run in `_simple_primary_switch`.

src/pg.py
```diff
@@ -794,18 +810,72 @@ def is_replaying_wal(self, check_time):
...
+            logging.warning('PostgreSQL is not in recovery. So we cannot enable wal receiver.')
+            return
+
+        if self._exec_query('SHOW primary_conninfo;').fetchone()[0] == '':
+            logging.info('ACTION. Enabling WAL receiver')
+            self._exec_query('ALTER SYSTEM RESET primary_conninfo;')
+            self._reload_conf()
+            logging.debug('WAL receiver enabled. Waiting for WAL receiver to start')
+            if not helpers.await_for(
+                self.check_walreceiver, self.config.wal_receiver_timeout * 2, 'WAL receiver is streaming'
```
mialinx, line 857: we don't need it in pgconsul.conf, as we can get it directly from postgresql
```
select setting::int/1000 from pg_settings where name = 'wal_receiver_timeout';
```

src/__init__.py
```diff
@@ -53,6 +53,7 @@ def read_config(filename=None, options=None):
             'use_replication_slots': 'no',
             'max_rewind_retries': 3,
             'postgres_timeout': 60,
+            'election_loser_timeout': 0,  # Timeout for election losers. For test purposes only.
```
mialinx, line 56: let's move this  optio to debug section and make more general.
Something like
```
debug:
   delay_name: 'election_looser',
   delay_value: 30,
```

by analogy with
```
if self.config.get('debug', 'failure_name', fallback=None) == name:
```

vicpopov (changes requested):

tests/steps/cluster.py
```diff
@@ -467,6 +467,13 @@ def step_container_is_replica(context, replica_name, primary_name):
     return assert_container_is_replica(context, replica_name, primary_name)
 
 
+@then('container "(?P<replica_name>[a-zA-Z0-9_-]+)" is a replica of container "(?P<primary_name>[a-zA-Z0-9_-]+)" and streaming')
+@helpers.retry_on_assert
+def step_container_is_replica_and_streaming(context, replica_name, primary_name):
+    return assert_container_is_replica(context, replica_name, primary_name) \
+        and step_container_is_in_quorum_group_and_streaming(context, replica_name)
```
vicpopov, line 474: No need to construct this long and, since both `assert_container_is_replica` and `step_container_is_in_quorum_group_and_streaming` are throwing AssertionError and do not return anything:
```suggestion
    assert_container_is_replica(context, replica_name, primary_name)
    step_container_is_in_quorum_group_and_streaming(context, replica_name)
```

vicpopov (changes requested):

vicpopov (approved):

vicpopov (approved):

PR #82, opened 2025-07-14
Set 10 seconds connection timeout in _check_primary_is_realy_dead

EinKrebs:


src/main.py
```diff
@@ -1775,7 +1775,7 @@ def _check_primary_is_really_dead(self, primary=None, check_primary=True):
             ensure_connect_primary = ''
 
         try:
-            conn = psycopg2.connect('host=%s %s %s' % (primary, append, ensure_connect_primary))
+            conn = psycopg2.connect('host=%s %s %s connect_timeout=10' % (primary, append, ensure_connect_primary))
```
vicpopov, line 1778: Isn't it a runtime configuration parameter?
Don't we have it already in `append`?
How does psycopg2 work with duplicate option `connect_timeout` if we have it everywhere?

PR #84, opened 2025-07-19
close on timeline missmatch even if maintenance is enabled

munakoiso:


tests/pgconsul.featureset
```diff
@@ -1,27 +1 @@
...
-features/async.feature
-features/kill_primary.feature
-features/kill_replica.feature
-features/kill_non_ha_replica.feature
-features/slot.feature
-features/coordinator.feature
-features/coordinator_fail.feature
-features/single_node.feature
-features/op_track.feature
 features/maintenance.feature
```
munakoiso, line 1: for faster debug in github actions

vicpopov: > for faster debug in github actions

Actually it is not used in github, we have matrix strategy, launching every feature in separate test: https://github.com/yandex/pgconsul/blob/main/.github/workflows/docker-behave.yml#L14

munakoiso: Oh... thanks

munakoiso:
finally:
      INFO:helpers:11:51:15: step_container_replication_state call: container "postgresql1" replication state is "async", while expected is "sync"
      https://github.com/yandex/pgconsul/actions/runs/16416069962/job/46382232807?pr=84

src/main.py
```diff
@@ -285,12 +285,16 @@ def start(self):
...
-    def update_maintenance_status(self, role, primary_fqdn):
+    def update_maintenance_status(self, role, primary_fqdn, zk_timeline, db_timeline):
         maintenance_status = self.zk.get(self.zk.MAINTENANCE_PATH)  # can be None, 'enable', 'disable'
 
         if maintenance_status == 'enable':
             # maintenance node exists with 'enable' value, we are in maintenance now
             self.is_in_maintenance = True
+            # verify if there was failover and our timeline is expired
+            if role == 'primary' and (zk_timeline is None or db_timeline is None or zk_timeline > db_timeline):
+                self.db.pgpooler('stop')
```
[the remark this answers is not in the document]
munakoiso: resolve_zk_primary_lock contains another check. Also it doesn't close master with empty ssn...
I think we should keep this check as simple as possible. But i'll add stop of archiving here

tests/features/maintenance.feature
```diff
@@ -302,3 +298,76 @@ Feature: Check maintenance mode
...
+          - client_hostname: pgconsul_postgresql2_1.pgconsul_pgconsul_net
+            state: streaming
+            sync_state: quorum
+          - client_hostname: pgconsul_postgresql3_1.pgconsul_pgconsul_net
+            state: streaming
+            sync_state: quorum
+        """
+        When we gracefully stop "pgconsul" in container "postgresql1"
+        When we disconnect from network container "postgresql1"
+        Then one of the containers "postgresql2,postgresql3" became a primary
```
vicpopov, line 354: NIT: It's very difficult to get without reading steps code, that this step is actually remembering container, which is later referenced as "remembered container".
I suggest adding "and we remember that container" to step text for be more clear.

PR #85, opened 2025-07-28
fix race in replication slot test

munakoiso:
host3 might be in archive recovery when we disconnecting host2 from network.
So, there might be no any moment for host1 to have two replication slots: we can drop replication slot for host2 before host3 starts streaming from anyone.


tests/features/cascade.feature
```diff
@@ -426,6 +426,13 @@ Feature: Check not HA hosts
           - slot_name: pgconsul_postgresql3_1_pgconsul_pgconsul_net
             slot_type: physical
         """
+        Then <lock_type> "<lock_host>" has following values for key "/pgconsul/postgresql/replics_info"
+        """
+          - client_hostname: pgconsul_postgresql2_1.pgconsul_pgconsul_net
```
vicpopov, line 431: `postgresql2` is primary, we can't have it as replica

munakoiso: nope, pg2 is replica

vicpopov (approved):
LGTM, waiting for tests

vicpopov (approved):

PR #91, opened 2025-08-15
remember last master in zk

munakoiso:


src/zk.py
```diff
@@ -531,7 +533,10 @@ def try_acquire_lock(self, lock_type=None, allow_queue=False, timeout=None, read
         Acquire lock (leader by default)
         """
         lock_type = lock_type or self.PRIMARY_LOCK_PATH
-        return self._acquire_lock(lock_type, allow_queue, timeout, read_lock=read_lock)
+        acquired = self._acquire_lock(lock_type, allow_queue, timeout, read_lock=read_lock)
+        if lock_type == self.PRIMARY_LOCK_PATH and acquired:
+            self.write(self.LAST_PRIMARY_PATH, self.get_lock_contender_name())
```
vicpopov, line 538: Why do you need to fetch contender name here, when you already know your hostname?

EinKrebs (approved):

PR #92, opened 2025-08-25
remember ssn of all hosts in zk

munakoiso:


src/zk.py
```diff
@@ -41,6 +41,10 @@ class Zookeeper(object):
     QUORUM_PATH = 'quorum'
     QUORUM_MEMBER_LOCK_PATH = f'{QUORUM_PATH}/members/%s'
 
+    SSN_BY_HOST_PATH = 'synchronous_standby_names_by_host'
```
vicpopov, line 44: We will have leftover garbage here if we remove host from cluster.
Can we place these debug data under `/alive/%s` path? I understand, not perfect, but could not find a better place.

And also I don't like `by_host` here.

Why can't we use one single path for it?
We only write it while we are primary, and keeping and logging SSN from host that was primary a year ago is below useful.

mialinx (approved):

PR #94, opened 2025-09-02
sleep before reinit zk connection

munakoiso:


rkhapov (approved):

vicpopov:
Are we sure this won't conflict with already existing backoff configuration for connections: https://github.com/munakoiso/pgconsul/blob/main/src/zk.py#L123?
Or maybe it will just not help?

PR #114, opened 2026-02-03
Reject switchover where no one holds the leader lock

munakoiso:


src/utils.py
```diff
@@ -173,10 +179,22 @@ def plan(self):
...
+            return self._zk.get_current_lock_holder(self._zk.PRIMARY_LOCK_PATH)
+
+        if primary is None:
+            primary = helpers.await_for_value(check_primary_lock_holder, self.timeout, "Primary holds the leader lock")
+        else:
+            self._log.info(f'Use {primary} as current primary')
+
+        if primary is None:
+            self._log.error('Switchover is impossible because no one holds the leader lock.')
+            sys.exit(1)
```
aak74: exit does not compatible with `get*` function

mialinx (dismissed):
LGTM

reshke (dismissed):
other than not-future-proof exit in _get_lock_owners its good as is

src/utils.py
```diff
@@ -171,17 +173,31 @@ def plan(self):
...
         """
-        Get leader and syncreplica lock owners, and timeline.
+        Plan switchover based on initial variables and current lock holders.
+        Return True if switchover is possible and False - if not
         """
+        def check_primary_lock_holder():
+            return self._zk.get_current_lock_holder(self._zk.PRIMARY_LOCK_PATH)
+
+        if self.primary is None:
+            self.primary = helpers.await_for_value(check_primary_lock_holder, self.timeout, "Primary holds the leader lock")
```
mialinx: Timeout may be long.. Should we fail fast here ?

mialinx (dismissed):

PR #140, opened 2026-03-27
Pg refactoring

mialinx:


src/main.py
```diff
@@ -2043,7 +2043,8 @@ def _do_primary_switchover(self, switchover_candidate, zk_state):
             return False
 
         # update replics info in ZK to avoid missguiding CLI
-        self._store_replics_info(self.db.get_state(), zk_state)
+        db_state['replics_info'] = self.db.get_replics_info('primary')
```
aak74, line 2024: We are already doing this here:
https://github.com/yandex/pgconsul/blob/main/src/main.py#L2178

src/pg.py
```diff
@@ -69,33 +80,23 @@ def __init__(self, config: PostgresConfig, plugins: PluginRunner, cmd_manager: C
...
-                cursor = self.conn_local.cursor()
-                cursor.execute('SELECT 1;')
-                return cursor
-            else:
-                raise RuntimeError('Local conn is dead')
-        except Exception:
-            for line in traceback.format_exc().split('\n'):
-                logging.debug(line.rstrip())
+        if self.conn_local is None:
             self.reconnect()
```
aak74, line 93: We don't want to check the connection's liveability? Will we always do reconnect?

src/pg.py
```diff
@@ -144,10 +157,9 @@ def _offline_detect_pgdata(self):
                 version, _, port, pgstate, _, pgdata, _ = row.split()
                 if port != need_port:
                     continue
-                if state.get('pg_version'):
+                if state:  # not empy
```
aak74: typo `not empty`?

src/pg.py
```diff
@@ -174,102 +186,86 @@ def reconnect(self):
...
+                    logging.warning('failed to close old connection: %s', err)
             self.conn_local = psycopg2.connect(self.config.conn_string)
             self.conn_local.autocommit = True
-
             self.role = self.get_role()
-            self.pg_version = self._get_pg_version()
             self.pgdata = self._get_pgdata_path()
-        except psycopg2.OperationalError:
-            logging.error('Could not connect to "%s".', self.config.conn_string)
+            self.terminal_state = True
```
aak74, line 199: If the postgres is dead, then we won't set terminal_state to True.

aak74 (changes requested):
Overall, I like it. There are a couple of questions

src/pg.py
mialinx: We will , see line 204.

src/main.py
mialinx: In line 2046  we update db_state with fresh value from PG.
In line 2178  we get this value from db_state dict, not from db.

Actualy we update it just for that place.

src/pg.py
mialinx: fixed

src/pg.py
mialinx: It's a tricky question...

In the target design, the reconnect function is designed to perform unconditional reconnections. It will be called only once (ok, twice) per cycle in `get_state`->`is_alive' to make sure that PG is really alive.
But in the target design, any (ok, most) exception would be caught at the top level, so we wouldn't need reconnections in the middle of an iteration.

In the current (interim) design, exceptions can be intercepted and handled by internal methods or wrappers, so we need to reconnect in the middle of an iteration. 

I have slightly changed the code, now any operational error will cause the connection to be disconnected, and it will be restored on the next request.

src/pg.py
```diff
@@ -216,47 +215,44 @@ def get_state(self):
...
+                #
+                # We ask health of PostgreSQL one more time since it could die
+                # while we were asking all other things here. It can lead to
+                # unpredictable results.
+                #
+                data['alive'] = self.is_alive()
         except Exception:
-            for line in traceback.format_exc().split('\n'):
-                logging.error(line.rstrip())
+            logging.exception("failed to get postgresql state")
```
munakoiso, line 253: Here we had and still have a problem: data may be not full if some query in the middle is failed.
Mb mark data inconsistent? We can not operate by this type of data

src/main.py
```diff
@@ -2043,7 +2043,8 @@ def _do_primary_switchover(self, switchover_candidate, zk_state):
...
+        self._store_replics_info(db_state, zk_state)
```
munakoiso, line 2025: looks like a bad idea
If you read state at the beginning of iteration - you should (choose one):
- use it til the end of iter
- update all state fields in the middle of iteration to be sure, that state is consistant.

in _store_replics_info -> write_host_stat for now you use replics_info that more fresh than wal_receiver.
not sure if its ok
not sure that we will not write some code, for which this behavior (with inconsistent state) will be unexpected.

src/pg.py
```diff
@@ -203,6 +203,18 @@ def reconnect(self):
             if any(e in str(err) for e in TRANSIENT_ERRORS):
                 self.terminal_state = False
 
+    def close(self):
+        """
+        Closes current connection in any state
+        """
+        logging.debug('Closing connection to GP')
```
munakoiso: pg

src/pg.py
```diff
@@ -69,34 +80,28 @@ def __init__(self, config: PostgresConfig, plugins: PluginRunner, cmd_manager: C
...
         self._plugins = plugins
         self._cmd_manager = cmd_manager
-
-        self.state: dict[str, object] = {}
-
         self.conn_local: psycopg2.extensions.connection | None = None
         self.role: str | None = None
         self.pgdata = ''
-        self.pg_version = None
+        # pg is either running or stopped, not starting ot stopping
```
munakoiso: ot -> or

src/pg.py
mialinx: Yes, you are right.
But i'm not sure at the moment we never use db_state fields whene pg is dead.
In `run_iteration` function we mostly rely on the `role` variable, so I suggest to get it from the `db_state` and clear it if last alive check failed.

src/main.py
mialinx: We can't rely on any kind of consistency between different keys in ZK (and even between different fields in `db_state`), because they are collected with different SQL queries. We should rely only on ZK locks and "happens after" relationships.

I agree, that updating db_state in the middle of itteration is not very nice, but trying to keep `db_state` atomic (semi-atomic actually) is a waste of effort.

src/pg.py
mialinx: thanks

munakoiso (approved):

aak74 (approved):

PR #144, opened 2026-04-13
MDB-44139: flap resistance in get_ha_hosts function

munakoiso:
We are currently non-atomically reading the list of ha_hosts from zk. So, if we get a zk exception when reading any ha node, we will immediately remove that node from the ha_hosts list. This may lead to incorrect cluster configuration (for example, single node mode may be enabled for a cluster with more than 2 ha nodes).

Now we will remove a host from ha_hosts only if its ha_node has been unavailable for a long time (default is 60 s).

### DIFF
--- a/src/zk.py
+++ b/src/zk.py
@@ -9,6 +9,7 @@ import os
 import traceback
 import time
 from random import uniform
+from .list_removal_strategy import DelayedListRemovalStrategy
 
 from kazoo.client import KazooClient, KazooState
 from kazoo.exceptions import LockTimeout, NoNodeError, KazooException, ConnectionClosedError, SessionExpiredError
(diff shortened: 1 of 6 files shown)


src/zk.py
```diff
@@ -545,6 +548,12 @@ def get_state(self):
         }
         data[self.LAST_PRIMARY_PATH] = self.get(self.LAST_PRIMARY_PATH)
         data['synchronous_standby_names'] = self._get_ssn_info()
+        catch_except = True
+        if self._ha_hosts is None:
+            catch_except = False
+            self._ha_hosts = []
+        ha_hosts_from_zk = self.get_ha_hosts(catch_except=catch_except, cached_value=False)
+        self._ha_hosts = self._ha_hosts_removal_strategy.get_hosts_to_keep(self._ha_hosts, ha_hosts_from_zk)
```
mialinx: It's better to fill and use cache in the same function - in `get_ha_hosts`.
Filling the cache in `get_state` creates implicit dependency: `get_ha_hosts` will not work without regular calls to `get_state`. "Action at a distance"

mialinx (changes requested):

mialinx (approved):

munakoiso:
https://github.com/yandex/pgconsul/pull/150 - better solution for stable branch
For main branch - current one is not good enough, we gonna rewrite it in MDB-44833

PR #145, opened 2026-04-13
cli should wait until old primary take health lock after switchover

mialinx:
in order to give it time to clean all temp zk keys like all_hosts/{fqdn}/tried_remaster

### DIFF
--- a/src/utils.py
+++ b/src/utils.py
@@ -116,6 +116,8 @@ class Switchover:
         state = self.state()
         self._log.debug('full state: %s', state)
         self._wait_for_replicas(ha_group, min_replicas)
+        if self._conf.getboolean('global', 'quorum_commit'):
+            self._wait_for_sync_group(ha_group, min_replicas)
         # We delete all zk states after switchover complete
         self._log.info('switchover finished, zk status "%(progress)s"', state)
         result = state['progress'] is None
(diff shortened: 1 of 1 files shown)


src/utils.py
```diff
@@ -277,6 +278,20 @@ def _wait_for_primary(self, timeout=None):
...
 
+    def _wait_for_old_primary(self, timeout=None):
+        """
+        Wait for primary to hold the lock
+        """
+        if timeout is None:
+            timeout = self.timeout
+
+        if not helpers.await_for(
+            lambda: self._zk.get_current_lock_holder(self._zk.get_host_alive_lock_path(self._plan['primary'])) == self._plan['primary'],
```
vicpopov: I'm afraid this won't work as intended, alive lock is not the desired indicator that old primary has performed rewind, because alive lock is not dropped in switchover scenario.
Alive lock is only dropped in `_zk_alive_refresh` [early in iteration](https://github.com/yandex/pgconsul/blob/stable/src/main.py#L362), but not during [rewind immediately after switchover](https://github.com/yandex/pgconsul/blob/stable/src/main.py#L2170)

I have two possible options in my mind:
1. Wait for quorum lock, which is obtained in final stages of replica iteration, when we are completely sure replica is fully functional: https://github.com/yandex/pgconsul/blob/stable/src/main.py#L1008
2. Extend method `_wait_for_replicas` with `old_primary` argument, and require it to become a replica in addition to `min_replicas` replics.

vicpopov (changes requested):

src/utils.py
vicpopov: > 2\. Extend method `_wait_for_replicas` with `old_primary` argument, and require it to become a replica in addition to `min_replicas` replics.

As we discussed, waiting for replication on master is not what we need, we need a proof of a complete replica iteration, which concludes with acquiring quorum lock.

src/utils.py
```diff
@@ -277,6 +307,21 @@ def _wait_for_primary(self, timeout=None):
...
+
+    def _wait_for_old_primary(self, timeout=None):
+        """
+        Wait for primary to hold the lock
+        """
+        if timeout is None:
+            timeout = self.timeout
+
+        if not helpers.await_for(
+            lambda: self._zk.get_current_lock_holder(self._zk.get_host_alive_lock_path(self._plan['primary'])) == self._plan['primary'],
```
vicpopov: For SingleSyncReplicationManager (quorum_commit not used) the desired lock would be `self._zk.SYNC_REPLICA_LOCK_PATH`: https://github.com/yandex/pgconsul/blob/stable/src/replication_manager.py#L253

vicpopov (changes requested):

src/utils.py
mialinx: > For SingleSyncReplicationManager (quorum_commit not used) the desired lock would be `self._zk.SYNC_REPLICA_LOCK_PATH`: https://github.com/yandex/pgconsul/blob/stable/src/replication_manager.py#L253

Yep.. but it we are going to get rid of single sync replication manager in the near future.
And it this lock may be held by any replica, not necessary old primary.

src/utils.py
```diff
@@ -265,6 +267,31 @@ def _wait_for_replicas(self, ha_group, min_replicas=None, timeout=None):
...
 
+    def _wait_for_sync_group(self, ha_group, min_replicas=None, timeout=None):
+        """
+        Wait until at least min_replicas HA replicas from ha_group hold their quorum member locks.
+        """
+        if timeout is None:
+            timeout = self.timeout
+        if min_replicas is None:
+            min_replicas = len(ha_group) - 1
+        min_replicas = min(min_replicas, len(ha_group))
```
vicpopov, line 278: This does not have any sense.
If `min_replicas` is above `len(ha_group) - 1`, it is an impossible configuration: we only have `len(ha_group) - 1` possible replicas in ha_group, and the last is primary.
Fortunately, we don't pass incorrect replicas parameter, so only a NIT.

vicpopov (approved):

mialinx:
will recreate over new stable (with tests)

PR #154, opened 2026-04-23
MDB-44933: delayed deletion of quorum hosts. short version for stable branch

munakoiso:


src/replication_manager.py
```diff
@@ -387,21 +389,36 @@ def update_replication_type(self, db_state, ha_replics):
...
                 logging.info('We should not change replication type here.')
                 return
-            if self.change_replication_to_quorum(quorum_hosts):
-                self._zk.write(self._zk.QUORUM_PATH, quorum_hosts, preproc=json.dumps)
+            if self.change_replication_to_quorum(new_quorum):
+                self._zk.write(self._zk.QUORUM_PATH, new_quorum, preproc=json.dumps)
                 logging.info('Turned synchronous replication ON.')
 
+    def _calculate_new_quorum(self, old_quorum, quorum_from_zk):
+        new_quorum = set(host for host in quorum_from_zk)
```
vicpopov: NIT:
Why this comprehension?
Won't `new_quorum = set(quorum_from_zk)` do the same?
We were casting this quorum from ZK to set perfectly fine in previous code: https://github.com/yandex/pgconsul/pull/154/changes#diff-85b283c50c288107310ead558a64a812605cafbb275800211c01540f2c8fd98fL398

tests/features/quorum_removal_delay.feature
```diff
@@ -0,0 +1,115 @@
...
+                role: replica
+            postgresql3:
+                role: replica
+        """
+        Then zookeeper "zookeeper1" has holder "pgconsul_postgresql1_1.pgconsul_pgconsul_net" for lock "/pgconsul/postgresql/leader"
+        Then container "postgresql2" is in quorum group
+        Then container "postgresql3" is in quorum group
+        
+        When we disconnect from network container "postgresql3"
+        And we wait "8.0" seconds
```
vicpopov: NIT:
I think we should demonstrate removal from quorum is not performed until timeout passes:
- sleep for 3 seconds
- check replica is still in quorum
- sleep for more 5 secoinds
- check replica is out of quorum

Actually, we can combine this scenario with previous one, and all quorum operations would be tested

tests/steps/cluster.py
```diff
@@ -974,6 +974,14 @@ def step_container_is_in_quorum_group_and_streaming(context, name):
     )
 
 
+@then('container "(?P<name>[a-zA-Z0-9_-]+)" is not in quorum group')
+@helpers.retry_on_assert
```
vicpopov: NIT:
This turns this step into a waiting one, which makes sleeps before it useless.
Maybe we should not add this decorator? Or remove sleeps in scenarios?

vicpopov (approved):

munakoiso:
not safe for stable branch:
planned maintenance of last replica will cause read-only of entire cluster

PR #155, opened 2026-04-24
MDB-42150: repl_mon for jepsen tests

munakoiso:


rkhapov (approved):

reshke (approved):

vicpopov (commented):
Why not for all tests? It would help behave tests too

munakoiso:
> Why not for all tests? It would help behave tests too

cause in behave we already have installed repl_mon

PR #157, opened 2026-04-28
MDB-37683: set ssn before promote

aak74:


src/replication_manager.py
```diff
@@ -6,15 +6,17 @@
 from .pg import Postgres
 from .list_removal_strategy import DelayedListRemovalStrategy
 from .replication_manager_factory import ReplicationManagerConfig
+from .ssn_manager import SsnManager
```
mialinx, line 9: There is no need in separate class - all logic may be placed in ReplicationManager.

src/ssn_manager.py
```diff
@@ -0,0 +1,121 @@
...
+    # Host-list assembly (pure / static)
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def build_replica_hosts_for_promote(
+        known_replicas: Iterable[str] | None,
+        extra_host: str | None = None,
+    ) -> list[str]:
+        """
+        Build the replica host list to pass to set_ssn_before_promote().
```
mialinx: Here is about 20 lines of comments for 4 lines of code.

I think this brings some problems to the code:
* Reading and understanding takes more time than reading actual code
* There are no guarantees that comments will correspond to the actual behavior (after few years)
* There is no tricky or not obvious logic here to be documented

I suggest to write documentation only in following cases:
* As short introduction for beginners (in separate markdown)
* To explain not obvious decisions (**why** this code works so, not **what** it does) 

In all other cases short and self-documenting code would be preferred.

src/replication_manager_factory.py
```diff
@@ -73,11 +73,14 @@ def create_replication_manager(config: RawConfigParser, db, zk):
     """
     # Import here to avoid circular dependencies and allow unit testing
     from .replication_manager import ReplicationManager
-    
+    from .ssn_manager import SsnManager
+
     replication_config = build_replication_manager_config(config)
-    
+    ssn_manager = SsnManager(db, zk)
+
```
mialinx: BTW, as removed SingleSynReplicationManager... do we still need factory to create ReplicationManager ?
May be simple constructor would be enough ?

src/main.py
```diff
@@ -873,6 +872,10 @@ def _accept_switchover(self, zk_state):
                 logging.warning('Some replicas are not streaming from the candidate...')
                 return False
 
+            if not self._replication_manager.set_ssn_before_promote(side_replicas, zk_state['lock_holder']):
```
mialinx: CRIT: side_replicas are not suitable for SSN, as it may contain cascade replicas, streaming from old primary.

tests/unit/test_replication_manager_ssn.py
```diff
@@ -0,0 +1,137 @@
...
+    ssn_manager = MagicMock(spec=SsnManager)
+    manager = ReplicationManager(_make_config(), db, zk, ssn_manager)
+    return manager, db, zk, ssn_manager
+
+
+# ---------------------------------------------------------------------------
+# ReplicationManager.set_ssn_before_promote
+# ---------------------------------------------------------------------------
+
+class TestSetSsnBeforePromote:
```
mialinx, line 48: During code-review I have to use simple search (ctrl+f) to jump between usages.
And usages in unit-test distract from main logic. 

Could we move unit test to separate files / directory to simplify navigation ?  (For all files)

src/ssn_manager.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Collects all hosts that will replicate from the new master after promote:
+        - known_replicas: hosts already streaming from the candidate (side replicas
+          during switchover) or HA replicas known from ZK (during failover)
+        - extra_host: an additional host to append, or None. During switchover this
+          is the current primary (lock_holder), which will start streaming from the
+          new master after the switchover completes.
+
+        extra_host is only included when known_replicas is non-empty.  In a
```
mialinx: NIT: Yep, that is the very important behavior.
But `build_replica_hosts_for_promote` sounds like pretty simple helper.

May be mooving this code to `set_ssn_before_promote` and renaming 
`known_replicas` -> `ha_replicas`, 
`extra_host` -> `old_primary`
would clarify our intentions ?

src/zk.py
```diff
@@ -721,17 +721,44 @@ def get_ssn_date_path(self, hostname=None):
...
-        if self.noexcept_get(value_path) != value:
-            self.noexcept_write(value_path, value, need_lock=False)
-            self.noexcept_write(date_path, time.time(), need_lock=False)
+                return True
+            except Exception:
+                for line in traceback.format_exc().split('\n'):
+                    logging.error(line.rstrip())
+                return None
+
+        retrying = helpers.get_exponentially_retrying(
```
mialinx: Writing ssn to ZK is only for debuggin. There is no need to retry it.

And futhermore, I would suggest to not write it to ZK at all.
Instead we could load current SSN as part of `db_state`  and write it to local log at every iteration.

@munakoiso what do you think about it ?

src/ssn_manager.py
```diff
@@ -0,0 +1,121 @@
...
+
+        The DB call is retried with exponential backoff for up to
+        ``self._apply_retry_timeout`` seconds.  The ZK write is retried
+        internally by :meth:`Zookeeper.write_ssn_on_changes`.
+
+        Logs 'ACTION. {start_msg}' before the first attempt and success_msg on
+        success.  Returns True on success, False if the DB timeout is exhausted.
+        """
+        logging.info(f'ACTION. {start_msg}')
+        retrying_db_call = helpers.get_exponentially_retrying(
```
mialinx: What is the scenario where simple ALTER without retries is not enough ?

Inner loops (especially with backoff) should be avoided, because pgconsul may stuck here and do not execute regular checks (as pg is alive). It may lead to to extra downtime in case of failover.

mialinx (changes requested):

src/replication_manager_factory.py
aak74: I would like to leave only logic in the ReplicationManager. I don't want to pull the RawConfigParser dependency into this class, and I also want to remove it from the Pgconsul class eventually.

src/ssn_manager.py
aak74: I agree with the renaming.
The separate build_replica_hosts_for_promote function is easy to test.

src/replication_manager.py
aak74: Maybe you're right. But it seems to me that a separate SsnManager is better tested and more understandable. The ReplicationManager is already quite large. But if you insist, I can combine it.

src/main.py
```diff
@@ -1711,6 +1718,10 @@ def _do_failover(self):
             self.zk.release_lock()
             return False
 
+        if not self._replication_manager.set_ssn_before_promote(self._get_ha_replics(), zk_state['lock_holder']):
```
mialinx: `self._get_ha_replics` will not work.
For example, we have cluster with 3 HA hosts: A and B and C.
`C` was a replica and is dead for a long time.
`A` was old primary and it failed (or lost network).
`B` is going to promote.
`_get_ha_replics` will return `(A, C)` for `B`, both of them are dead, B will stuck after promote.

I guess we should use here the same logic as on primary in `update_replication_type`
https://github.com/yandex/pgconsul/blob/main/src/replication_manager.py#L201
The simplest way to do this - use hosts written in`self._zk.QUORUM_PATH`.

src/main.py
```diff
@@ -1697,7 +1704,7 @@ def _accept_failover(self, switchover_in_progress=False):
             logging.exception('Unexpected error while trying to do failover. Exiting.')
             sys.exit(1)
 
-    def _do_failover(self):
+    def _do_failover(self, zk_state):
```
mialinx: I think we don't need to pass `zk_state` through all the stack to get `zk_state['lock_holder']`.
Because absence of this lock is the condition to start failover procedure: https://github.com/yandex/pgconsul/blob/main/src/main.py#L956

src/main.py
```diff
@@ -872,6 +871,10 @@ def _accept_switchover(self, zk_state):
                 logging.warning('Some replicas are not streaming from the candidate...')
                 return False
 
+            if not self._replication_manager.set_ssn_before_promote(self._get_ha_replics(), zk_state['lock_holder']):
```
mialinx: 1) The same problem.. will not work, because `_get_ha_replics` may return dead hosts.

2) At this moment `zk_state['lock_holder']` contains old primary.
It doesn't make sence to add it to the SSN, because it will not join replication until rewind.
It *may* work if `_get_ha_replics` contains at alive hosts, but otherwise new primary will stuck.

I suppose it's better not to try to add guaranteed dead hosts to the SSN.


@munakoiso , your opinion ?

src/replication_manager.py
mialinx: @munakoiso , you opinion ?

mialinx (changes requested):

src/main.py
aak74: In the described scenario, the following will happen:
- Before promote, we will set `SSN = ANY 1(A,C)`. - This will prevent writing to the new primary.
- In the next iteration, `replication type: async'.
- The cluster will be available for recording

It seems to me that this is a reasonable compromise between data security, the unavailability of the cluster for writing, and the simplicity of the code.

For clarity, I added this scenario to the test.
https://github.com/yandex/pgconsul/pull/157/changes#diff-457e8650761ba8767a633fe9221bdba71af1d7d3f38a36d7304659e59c24231aR160

mialinx (approved):
LGTM

PR #158, opened 2026-04-30
MDB-42341: handle exception in pg_wal_replay_resume + fix race condition in cascade

aak74:
Wrap WAL replay resume logic in try/except to prevent unhandled exceptions from crashing the iteration. On failure, a warning is logged with traceback and the operation is retried on next cycle.

Fix race condition in cascade.feature

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,8 +30,10 @@ src/                    # Main source code (pgconsul package)
 ├── exceptions.py       # Custom exceptions
 ├── list_removal_strategy.py       # Quorum list removal strategy
 ├── ssn_manager.py      # SSN (Sync Standby Names) management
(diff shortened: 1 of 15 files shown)


tests/features/cascade.feature
vicpopov, line 383: The problem with this scenario is actually here: recovery_conf generation command generates primary_conninfo without slots requirement, so replicas connect to master even before slots are created.
This ruins the scenario which otherwise would work.

vicpopov (changes requested):

tests/features/cascade.feature
aak74: Scenario assumes that slots will be created automatically. That is, the command looks like the right one.
```gherkin
Scenario Outline: Replication slots created automatically
```

aak74:
Fixed here #208

PR #162, opened 2026-05-09
Optimize Docker infrastructure, speed up container startup and improve test execution logging

aak74:
- Speed up container startup on MacOS by 2x (from ~180s to ~80s)
- Add make and build-essential to base image for compilation
- Generate SSL certificates at image build


aak74:
It seems that the problem only occurs on MacOS when using colima & aarch64. No problems have been observed on x86_64.

PR #166, opened 2026-06-19
Adapt MDB-37683 for stable-branch: set ssn before promote.

munakoiso:


src/main.py
```diff
@@ -1737,6 +1737,23 @@ def _do_failover(self):
             self.zk.release_lock()
             return False
 
+        change_replication = self.config.getboolean('primary', 'change_replication_type')
+        if change_replication:
+            quorum = self.zk.get(self.zk.QUORUM_PATH, preproc=helpers.load_json_or_default)
```
munakoiso: Here I am using a QUORUM_PATH, which is not suitable for SingleSyncReplicationManager. But it is deprecated

src/main.py
```diff
@@ -1737,6 +1737,23 @@ def _do_failover(self):
...
+                quorum = []
+            my_hostname = helpers.get_hostname()
+            quorum = [e for e in quorum if e != my_hostname]
+            if old_master:
+                quorum += [old_master]
+            quorum = sorted(set(quorum))
+            # We will update ssn in stable branch only in happy path.
+            # Because if it fails - it seems it would be better to leave everything as it is than to cancel the failover.
+            try:
+                self._replication_manager.update_replication_type(None, None, force_quorum=quorum)
```
aak74: I noticed that there's no validation for empty quorum before calling `update_replication_type()`. In scenario 2 of the test suite, when all HA replicas are unavailable, the `quorum` list ends up empty.

When an empty quorum is passed to `update_replication_type()`, it tries to set `synchronous_standby_names = 'ANY 0()'`, which causes a PostgreSQL syntax error:

```
psycopg2.errors.SyntaxError: invalid value for parameter "synchronous_standby_names": "ANY 0()"
DETAIL:  syntax error at or near ")"
```

PostgreSQL doesn't support `ANY 0()` or `FIRST 0()` - the number must be ≥ 1. This causes the failover to fail with a critical error.

It would be good to add validation to check if the quorum is empty before calling `update_replication_type()`. If the quorum is empty, we could either skip the SSN update (and log a warning) or abort the failover with a clear error message.

src/replication_manager.py
```diff
@@ -365,12 +365,18 @@ def init_zk(self):
             return False
         return True
 
-    def update_replication_type(self, db_state, ha_replics):
+    def update_replication_type(self, db_state, ha_replics, force_quorum=None):
```
aak74: I have a suggestion regarding the API design here. The parameter name force_quorum suggests a boolean value (force or not force), but we're actually passing a list of hosts as the quorum.

This creates some confusion about the parameter's purpose. The current approach works, but it might be less maintainable and harder to understand for other developers.

src/main.py
```diff
@@ -1737,6 +1737,23 @@ def _do_failover(self):
...
+        change_replication = self.config.getboolean('primary', 'change_replication_type')
+        if change_replication:
+            quorum = self.zk.get(self.zk.QUORUM_PATH, preproc=helpers.load_json_or_default)
+            if quorum is None:
+                quorum = []
+            my_hostname = helpers.get_hostname()
+            quorum = [e for e in quorum if e != my_hostname]
+            if old_master:
+                quorum += [old_master]
+            quorum = sorted(set(quorum))
```
aak74, line 1751: I have a concern about the quorum formation logic here. During `switchover`, the losing replica is removed from `QUORUM_PATH` to preserve data consistency. This means that when we read `QUORUM_PATH` at line 1742, we're getting a list that doesn't include the losing replica.

The issue is that we then add `old_master` to the quorum (line 1748), but at this point the old master is already in shutdown and cannot acknowledge transactions. Meanwhile, available replicas (like side replicas from the `switchover`) are not being included in the quorum.

This creates a situation where SSN is set with an unavailable old_master, which doesn't actually prevent data loss. The available replica that could acknowledge transactions is being ignored.

Would it be better to include side replicas from the `switchover` in the quorum formation? This would ensure that SSN is set with replicas that are actually available and can acknowledge transactions.

aak74 (dismissed):

src/main.py
aak74: I used test: https://github.com/yandex/pgconsul/blob/main/tests/features/ssn_before_promote.feature

munakoiso: I will add validation to prevent useless error, but it's not "causes the failover to fail with a critical error.", because update_replication_type wrapped with try-except block.

Also, empty quorum is almost impossible situation for normal failover/switchover:

In switchover old master alive and releases lock before the moment when new master calls _do_failover function - so no one can make quorum node in zk empty.

In failover quorum of hosts should vote before _do_failover call. Only one possibility to quorum to be empty is race like:
- old master get stuck somewhere after checking leader lock for a while 
- quorum waits for timeout, votes, choose new master and right after that
- old master continues iteration and drops all hosts from quorum. 

In stable branch we will ignore that race. As i notice in comment - it's "best-effort" fix
In main branch there is no such race, because of unavailability timeout before removing replica from quorum.

src/main.py
munakoiso: New master can commit with all side replicas (which should be in QUORUM-node. read previous comment)
Quorum of replicas from QUORUM-node should be available where we do swithover/failover. If not - we can not commit anything or we potentially loosing data.

Here i'm adding old master into QUORUM-node, but in fact - old master is a part of quorum hosts.
We not writing current master into QUORUM-node. Here i'm solving this problem

aak74: I'm trying to say that the problem isn't adding the old master, but the lack of a prompt before the switchover.

munakoiso: My bad, there is change_replication_to_sync_host call in switchover

fixed, waiting for tests result

aak74:
LGTM

aak74 (approved):

PR #172, opened 2026-06-25
Remove misleading zk reconnect message

aak74:
On the first ZK reconnect attempt (`_failed_inits_count == 0`), the code logged a misleading "attempt #0 failed" message and applied a backoff delay, even though no failure had occurred yet.

Now the first attempt is logged at `debug` level with no delay. Backoff is applied only from the second attempt onward.

### DIFF
--- a/src/zk.py
+++ b/src/zk.py
@@ -333,7 +333,7 @@ class Zookeeper(object):
             self._failed_inits_count,
             sleep_time,
             max_sleep,
-            self._max_delay_on_reinit,
+            self._max_delay_on_reinit
         )
         time.sleep(sleep_time)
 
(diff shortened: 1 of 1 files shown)


munakoiso (dismissed):

src/zk.py
```diff
@@ -327,15 +327,18 @@ def _sleep_before_reconnect(self):
...
-        logging.warning(
-            "ZK reconnection attempt #%d failed. Applying exponential backoff: sleeping %.2fs "
-            "(max possible: %.2fs, configured max: %ds)",
-            self._failed_inits_count,
-            sleep_time,
-            max_sleep,
-            self._max_delay_on_reinit
-        )
-        time.sleep(sleep_time)
+        if self._failed_inits_count > 0:
```
fedusia: Let use early return on first run:
```
if self._failed_inits_count == 0:
        logging.debug("Reconnecting to ZK (first attempt, no delay)")
        return
```
We don't waste time on calculations or sleep, but exit the method immediately (writing only a quick debug log). This makes sense, since at the first connection error, it's worth trying to reconnect immediately.

then you can remove the nesting (if/else) and simply call logging
The code has become more readable: unnecessary nesting of conditions has been removed.

fedusia (approved):

PR #173, opened 2026-06-26
MDB-46149: prevent restart of overloaded but alive postgres

aak74:
Add **pg_conn_failure_grace_period** (default 0s): when a connection
timeout occurs but systemctl reports postgres as running, pgconsul
now waits out the grace period before forcing a restart. Also adds
exponential connect_timeout backoff (1→2→4→8→10s) and diagnostics
logging on timeout (pg_status, loadavg, cpu_count, PSI metrics).

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -60,8 +60,9 @@ class pgconsul(object):
         self._debug_counters: dict[str, int] = {}
         self.last_zk_host_stat_write: float = 0
         self._replication_manager = self._get_repllication_manager()
-        # Number of consecutive iterations where Postgres connection timed out.
-        # Used to decide when to force a restart (see run_iteration).
+        # Restart threshold counter: consecutive iterations where Postgres
+        # connection timed out. Used to decide when to force a restart.
+        # Independent from _conn_timeout_count in pg.py (which drives backoff).
         self._pg_timeout_count: int = 0
         self._max_pg_timeouts: int = self.config.getint('global', 'max_conn_timeouts_before_restart')
         if self._max_pg_timeouts < 1:
@@ -380,6 +381,9 @@ class pgconsul(object):
             self._pg_timeout_count += 1
             pg_status = self.db.get_postgresql_status()
             role = None
+            # True: skip the "waiting for startup/shutdown" branch in dead_iter().
+            # Below threshold dead_iter() returns early; at threshold we want to
+            # proceed to restart without waiting.
             terminal_state = True
             db_state = {
                 'alive': False,
--- a/src/main.py
+++ b/src/main.py
@@ -18,7 +18,7 @@ import psycopg2
 
 from configparser import RawConfigParser
 
-from . import helpers, sdnotify
+from . import exceptions, helpers, sdnotify
 from .command_manager import CommandManager, Commands
 from .failover_election import ElectionError, FailoverElection
 from .helpers import IterationTimer, get_hostname, register_sigterm_handler, should_run
@@ -60,6 +60,10 @@ class pgconsul(object):
         self._debug_counters: dict[str, int] = {}
         self.last_zk_host_stat_write: float = 0
         self._replication_manager = self._get_repllication_manager()
+        # Number of consecutive iterations where Postgres connection timed out.
+        # Used to decide when to force a restart (see run_iteration).
+        self._pg_timeout_count: int = 0
+        self._max_pg_timeouts: int = self.config.getint('global', 'max_conn_timeouts_before_restart')
 
     def _get_repllication_manager(self) -> ReplicationManager:
         if self.config.getboolean('global', 'quorum_commit'):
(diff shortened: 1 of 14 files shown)


src/pg.py
```diff
@@ -75,26 +78,44 @@ def __init__(self, config: PostgresConfig, plugins: PluginRunner, cmd_manager: C
         self.role: str | None = None
         self.pgdata = ''
         self.pg_version = None
+        # Backoff counter for connect_timeout (1→2→4→8→10s). Reset on success.
+        # Independent from _pg_timeout_count in main.py (restart threshold).
```
munakoiso: there is no _pg_timeout_count is main.py

src/pg.py
```diff
@@ -75,26 +78,44 @@ def __init__(self, config: PostgresConfig, plugins: PluginRunner, cmd_manager: C
...
+        self._conn_timeout_count = 0
+        self._base_conn_string = self._strip_connect_timeout(config.conn_string)
         self._offline_detect_pgdata()
         self.reconnect()
 
     def _create_cursor(self):
+        def _open_cursor():
+            if self.conn_local is None:
+                raise RuntimeError('Local conn is dead in _open_cursor()')
```
munakoiso, line 90: this code is unreachable - in both _open_cursor calls self.conn_local is True => is not None

src/pg.py
munakoiso: mb it will be better to use logging.debug here?
Error is "Could not connect to...". All error lines is just debug information, not error itself

src/main.py
```diff
@@ -355,13 +359,34 @@ def _update_replication_on_maintenance_enter(self):
...
+            role = None
+            # True: skip the "waiting for startup/shutdown" branch in dead_iter().
+            # Below threshold dead_iter() returns early; at threshold we want to
+            # proceed to restart without waiting.
+            terminal_state = True
+            db_state = {
+                'alive': False,
+                'running': pg_running,
+                'prev_state': self.db.get_cached_state(),
+                'connection_timeout': True,
```
munakoiso: connection_timeout looks like variable, that contains timeout itself, not flag that "previous connection attempt was timed out". 
Mb connection_timed_out is better?

src/pg.py
```diff
@@ -168,6 +189,50 @@ def _drop_replication_slot(self, slot_name):
...
+        return min(
+            self.LOCAL_CONNECT_TIMEOUT_INITIAL * (2 ** self._conn_timeout_count),
+            self.LOCAL_CONNECT_TIMEOUT_MAX,
+        )
+
+    def _log_connection_failure_diagnostics(self, connect_timeout: int):
+        """Log system diagnostics when PostgreSQL connection fails with timeout."""
+        try:
+            pg_status = self.get_postgresql_status()
+            logging.warning('Connection timeout diagnostics: pg_status=%s', pg_status)
```
munakoiso: pg_status=0 is unreadable
lets rewrite it with "postgres is [not] alive (pg_status return code [not] 0)"

tests/features/overloaded_primary.feature
```diff
@@ -0,0 +1,235 @@
...
+                    quorum_commit: 'yes'
+                    pg_conn_failure_grace_period: 7
+                primary:
+                    change_replication_type: 'yes'
+                    primary_switch_checks: 1
+                replica:
+                    allow_potential_data_loss: 'no'
+                    primary_switch_checks: 1
+                    min_failover_timeout: 120
+                    primary_unavailability_timeout: 2
```
munakoiso, line 114: is 2s always enough for pg1 to became alive after restart?
if not - this test will be flaky

tests/features/overloaded_primary.feature
```diff
@@ -0,0 +1,235 @@
...
+        Then postgresql in container "postgresql1" was restarted
+        Then zookeeper "zookeeper1" has holder "pgconsul_postgresql1_1.pgconsul_pgconsul_net" for lock "/pgconsul/postgresql/leader"
+        And container "postgresql1" became a primary
+        And "pgbouncer" is running in container "postgresql1"
+        And container "postgresql2" is a replica of container "postgresql1"
+        And container "postgresql3" is a replica of container "postgresql1"
+
+    @skipping_restart @skipping_restart_replica
+    Scenario: Overloaded replica is not restarted while systemctl reports it running using default pg_conn_failure_grace_period
+        # pg_conn_failure_grace_period is NOT set in config — verifies the default value (30.0s) is applied.
```
munakoiso: default pg_conn_failure_grace_period is 0 in this pr

tests/features/overloaded_primary.feature
```diff
@@ -0,0 +1,235 @@
...
+        And container "postgresql2" is a replica of container "postgresql1"
+        And container "postgresql3" is a replica of container "postgresql1"
+        And "pgbouncer" is running in container "postgresql2"
+
+        When we remember postgresql start time in container "postgresql2"
+
+        # SIGSTOP simulates overload on replica: process alive (systemctl=running) but connection times out.
+        When we kill "postgres" in container "postgresql2" with signal "STOP"
+
+        # pg_conn_failure_grace_period=30s (default). 15s elapsed < 30s → Skipping.
```
munakoiso: and here - wrong default

munakoiso (changes requested):

src/pg.py
aak74: Yes. But:
1. The types need to be satisfied.
2. I don't want to rewrite the inner function when the call sites change.

munakoiso (approved):

PR #175, opened 2026-07-01
MDB-42077: remove the plugin mechanism

aak74:


src/pgbouncer.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, enabled=False):
+        self._enabled = enabled
+
+    def before_populate_recovery_conf(self, primary_host):
+        """
+        Update pgbouncer.ini to point to the new primary host.
+        Called before generating recovery.conf.
+        """
+        if self._enabled:
+            cmd = 'sudo sed -i /etc/pgbouncer/pgbouncer.ini -e "/^* = /s/host=.*$/host=' + primary_host + ' port=6432/"'
```
mialinx: It's a hardcode.  6432 port is our proprietary hack not used in open-source.
Let's try to figure out why did we need this plugin at all.
I guess it's legacy and we can get rid of it completely.

src/wal_uploader.py
```diff
@@ -1,16 +1,42 @@
...
+        Args:
+            config: Configuration object with plugins settings
+            conn: Database connection
+        """
+        self._config = config
+        self._wals_to_upload = 20
+        if hasattr(config, 'plugins') and config.plugins.get('wals_to_upload'):
+            self._wals_to_upload = int(config.plugins.get('wals_to_upload', 20))
+
+    def after_promote(self, conn):
```
mialinx: We just moved this code to another place..
I suggest to place this code in `pg.py` as it works with postgres data and name `def upload_wal` to clarify our intentions in https://github.com/yandex/pgconsul/pull/175/changes#diff-1dd21d0846b6c985aaf231799772c4797a69429020c8d3781caf8e285d35db43R523

src/__init__.py
```diff
@@ -61,6 +60,7 @@ def read_config(filename=None, options=None):
             'priority': 0,
             'update_prio_in_zk': 'yes',
             'standalone_pooler': 'yes',
+            'manage_pgbouncer_config': 'no',
```
mialinx: do we still need this option ?

mialinx (changes requested):

src/wal_uploader.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        Initialize WAL uploader.
+
+        Args:
+            config: Configuration object with plugins settings
+            conn: Database connection
+        """
+        self._config = config
+        self._wals_to_upload = 20
+        if hasattr(config, 'plugins') and config.plugins.get('wals_to_upload'):
```
vicpopov: You have dropped plugins from config: https://github.com/yandex/pgconsul/pull/175/changes#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312L100 and yet continue to read its contents and rely on it.
You should move to new location of this setting with fallback to old location, because noone would remember to update config simultaneously with pgconsul package.

src/pg.py
```diff
@@ -546,16 +542,80 @@ def _wait_for_primary_role(self):
...
+            for i, wal in enumerate(wals_to_upload_list, 1):
+                path = '{pgdata}/pg_wal/{wal}'.format(pgdata=pgdata, wal=wal)
+                cmd = archive_command.replace('%p', path).replace('%f', wal)
+                logging.info(f"[{i}/{len(wals_to_upload_list)}] Uploading WAL: {wal}")
+                helpers.subprocess_call(cmd)
+
+            logging.info("WAL upload completed successfully")
+        except Exception as error_message:
+            logging.error(f"WAL upload failed with error: {error_message}", exc_info=True)
+
```
vicpopov, line 604: Same code in `wal_uploader.py`. What is it for here?

vicpopov (changes requested):

src/pg.py
aak74: This is a request from @mialinx
wal_uploader.py will be deleted.

src/wal_uploader.py
aak74: fixed

src/pg.py
aak74: fixed

src/pg.py
```diff
@@ -75,11 +74,11 @@ class Postgres(object):
     DISABLED_ARCHIVE_COMMAND = '/bin/false'
     DISABLED_RESTORE_COMMAND = '/bin/false'
 
-    def __init__(self, config: PostgresConfig, plugins: PluginRunner, cmd_manager: CommandManager):
+    def __init__(self, config: PostgresConfig, cmd_manager: CommandManager):
         self.config = config
-        self._plugins = plugins
         self._cmd_manager = cmd_manager
         self.conn_local: psycopg2.extensions.connection | None = None
+        self._wals_to_upload = config.wals_to_upload
```
mialinx: We already have `self.config`, so we may just use `self.config.wals_to_upload`

src/pg.py
```diff
@@ -546,16 +542,80 @@ def _wait_for_primary_role(self):
...
+
+                cur.execute("SHOW archive_command")
+                archive_command = cur.fetchone()[0]
+                logging.debug(f"Original archive_command: {archive_command}")
+
+                # wal-g upload in parallel by default
+                if 'envdir' in archive_command:
+                    archive_command = "/usr/bin/envdir /etc/wal-g/envdir sh -c 'WALG_UPLOAD_CONCURRENCY=1 {}'".format(
+                        archive_command.replace('/usr/bin/envdir /etc/wal-g/envdir ', '')
+                    )
```
mialinx: Hardcode.   
Is this hack still needed in our setup ? 
We may just drop it (as pgbouncer related code) or may override `archive_command` in  pgconsul config.

mialinx (changes requested):

mialinx (approved):
LGTM

PR #180, opened 2026-07-07
MDB-41956: split Zookeeper into ZkClient and high-level facade

aak74:
Separate low-level KazooClient handling (connection lifecycle,
reconnect with backoff, domain exceptions, LockHandle) into ZkClient
and turn Zookeeper into a typed facade with semantic methods for every
ZK path operation. Update all call sites and add unit tests.

### DIFF
--- a/src/zk_client.py
+++ b/src/zk_client.py
@@ -1,0 +1,450 @@
+# encoding: utf-8
+"""
+ZkClient module. Low-level KazooClient wrapper for ZooKeeper connection management.
+"""
+
+import logging
+import os
+import time
+from enum import Enum
+from random import uniform
+from typing import Callable, List, Optional
+
+from kazoo.client import KazooClient, KazooState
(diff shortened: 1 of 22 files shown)


src/zk_client.py
```diff
@@ -0,0 +1,454 @@
...
+_KAZOO_STATE_MAP = {
+    KazooState.CONNECTED: ZkConnectionState.CONNECTED,
+    KazooState.SUSPENDED: ZkConnectionState.SUSPENDED,
+    KazooState.LOST: ZkConnectionState.LOST,
+}
+
+
+# === Lock handle ===
+
+class LockHandle:
```
munakoiso, line 67: Mb we should check NoNodeError separately from KazooException here?

aak74: fixed

munakoiso (approved):

PR #182, opened 2026-07-09
MDB-47201: handle connection errors gracefully in lock operations for ClickHouse Keeper compatibility (main)

aak74:


aak74:
There should be no code fixes for ClickHouse Keeper to work.

PR #183, opened 2026-07-09
MDB-47201: try using clickhouse-keeper instead of zookeeper (stable)

aak74:
For testing purposes only. No need to merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,7 @@ build_pgconsul:
 		--label pgconsul_tests
 
 jepsen_test:
+	docker compose -p $(PROJECT) -f jepsen-compose.yml build zookeeper1 zookeeper2 zookeeper3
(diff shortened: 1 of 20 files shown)


src/main.py
```diff
@@ -1413,7 +1413,10 @@ def _acquire_replication_source_slot_lock(self, source):
             )
         if source:
             # And acquire lock (then new_primary will create replication slot)
-            self.zk.acquire_lock(os.path.join(self.zk.HOST_REPLICATION_SOURCES, source), read_lock=True)
```
mialinx, line 1416: CRIT! Looks not fine.
Do CH and ZK have different protocols here ?
We expect that changing DCS implementation will not affect application logic at all, especially in so many places.

mialinx (changes requested):

PR #187, opened 2026-07-14
MDB-34331: make pgconsul a real daemon

mialinx:
* don't exit in the normal situation
* just skip iterations if rewind_failed flag set
* removed initial checks that duplicate run_iteration logic


src/main.py
```diff
@@ -323,6 +311,10 @@ def _update_replication_on_maintenance_enter(self):
     def run_iteration(self, my_prio):
         logging.info('Start iteration on host: %s', helpers.get_hostname())
         timer = IterationTimer()
+        if self.is_rewind_flag_set():
+            logging.error('Rewind fail flag is set, skipping iteration. Remove %s to resume.', self._rewind_flag_path())
```
vicpopov, line 317: This is a breaking behaviour, our current resetup routine depends on removing rewind fail flag first, then running highstate, which delays launch of pgconsul until replica is already streaming.
And we have discussed it in https://st.yandex-team.ru/MDB-31734#67113e3a2eb0d759a8157d9a, and you rejected that approach, but it could help here.

vicpopov (changes requested):

src/main.py
mialinx: Yes, this PR is created exactly to simplify resetup.
I'm going to keep rewind_failed flag during resetup and clean it only after HS completed sucessfully.
There is TODO for resetup fix: https://st.yandex-team.ru/MDB-47722

aak74 (approved):
LGTM

PR #188, opened 2026-07-16
MDB-47745: run behave tests with high network latency

munakoiso:


src/main.py
```diff
@@ -293,6 +293,8 @@ def update_maintenance_status(self, db_state, zk_state):
         elif maintenance_status is None:
             # maintenance node doesn't exists, we are not in maintenance mode
             self.is_in_maintenance = False
+        else:
+            logging.error('ALARM: unexpected maintenance status, %s', maintenance_status)
```
aak74, line 297: If the node was in maintenance,`is_in_maintenance` stays `True` - it may need to be explicitly reset.

tests/steps/latency.py
```diff
@@ -0,0 +1,144 @@
...
+        if not entry:
+            continue
+        pair_str, ms_str = entry.split(':')
+        dcs = pair_str.strip().split('-')
+        if len(dcs) != 2:
+            raise ValueError(
+                f"Invalid DC pair '{pair_str}', expected format 'dcA-dcB'"
+            )
+        dc_a, dc_b = dcs[0].strip(), dcs[1].strip()
+        delays[(dc_a, dc_b)] = int(ms_str.strip())
```
aak74: We may need a few more validations:
```py
        try:
            delay_ms = int(ms_str.strip())
        except ValueError:
            raise ValueError(f"Invalid delay '{ms_str}' in entry '{entry}', expected integer ms")
        if delay_ms < 0:
            raise ValueError(f"Negative delay {delay_ms} in entry '{entry}'")
        delays[(dc_a, dc_b)] = delay_ms
```

tests/steps/latency.py
```diff
@@ -0,0 +1,144 @@
...
+    manager = NetworkLatencyManager(config)
+    manager.apply(dc_map)
+    context.latency_manager = manager
+    LOG.info('Network latency applied to nodes: %s', all_nodes)
+
+
+def cleanup_latency(context):
+    """Remove latency rules during scenario teardown.
+
+    Called from after_scenario().  Since containers are killed anyway,
```
aak74: It looks like this method doesn't provide value in the current setup:

1. It is called in [`after_all`](tests/environment.py:78), after **all** scenarios.
2. In [`after_scenario`](tests/environment.py:117) every container is already
   killed and removed (`container.remove(v=True, force=True)` +
   `context.containers.clear()`).
3. `tc` rules are applied **inside the container's network namespace**. When
   `container.remove` runs, the netns is destroyed along with all rules.
4. By `after_all` there are no containers → `latency_manager.remove()` is a
   no-op (or an exception swallowed by `try/except`).

Even moving it to `after_scenario` **before** killing containers probably
wouldn't help — the container is killed immediately after, and tc rules die
with the netns.

I might be missing a case where `NetworkLatencyManager.remove()` cleans up
something outside containers (e.g., host-side qdisc on the Docker bridge). If
so, it would be great to note that in a comment and fix the docstring
(`after_all`, not `after_scenario`). If not, consider removing the method and
the call in `after_all`.

tests/steps/zk.py
```diff
@@ -191,15 +191,20 @@ def step_zk_set_value(context, value, key, name):
...
+        encoded = value.encode()
         try:
-            zk.set(key, value.encode())
+            zk.set(key, encoded)
         except kazoo.exceptions.NoNodeError:
-            zk.create(key, value.encode())
+            try:
+                zk.create(key, encoded, makepath=True)
+            except kazoo.exceptions.NodeExistsError:
+                zk.set(key, encoded)
```
aak74: This code overlaps with `zk_client.write`. It might be worth extracting into a shared helper to avoid
duplication.

Something like `helpers.write_zk_value(zk, key, value)` and reuse in
both places.

aak74 (changes requested):

src/main.py
munakoiso: I think that if the value of the node is unknown, we should not do anything so that there is no possibility of harm

aak74 (approved):
LGTM

PR #199, opened 2026-08-02
MDB-48301: Wait for walreceiver stop before failover election

munakoiso:
## Problem

During failover a replica clears `primary_conninfo` and reloads so that it stops receiving WAL from the old primary before it reads LSN and votes. That disable is applied by Postgres **startup asynchronously**. Reload returning does **not** mean walreceiver is already gone.

Until startup actually stops the receiver, the replica can still:
- receive WAL onto disk;
- send sync acknowledgements to the old primary.

`pg_wal_replay_pause()` does not close this window: it only stops **replay**, not receive/ack. So a replica can vote with an LSN taken while walreceiver is still alive, and the resurrected old primary can still get sync commits acknowledged after voting has started — a split-brain / divergent-write risk.

## Fix

Before entering the election:
1. set `primary_conninfo` to empty and reload (if needed);
2. wait until `pg_stat_wal_receiver` is empty, up to `replica.walreceiver_disable_timeout` (default 5s);
3. if the receiver does not stop in time — abort this failover attempt;
4. only then read LSN and participate in the election.

`pg_wal_replay_pause` was removed from this path as ineffective for the problem above.

## Test

New scenario in `failover_with_network_inconsistency.feature` forces the worst case:
- replicas enter failover and hit a debug sleep before disable;
- old primary is started again and walreceivers reconnect;
- replica `postgres: startup` is SIGSTOP’d (auto-CONT after N seconds), so disable cannot take effect while frozen;
- after votes appear in ZK, a sync `CREATE TABLE` on the old primary must **not** complete (app-side deadline; `statement_timeout` does not cancel SyncRep waits).

With the fix, replicas do not vote until the receiver is really stopped (after CONT / successful wait). Without the wait, they can vote while the receiver is still acking and the write can succeed.

Debug-only knobs for the scenario: `debug.pre_pause_sleep`, `debug.election_lsn_read_sleep`.

New scenario green with this change; red if wait-for-stop is removed

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -108,6 +108,9 @@ def read_config(filename=None, options=None):
             'primary_switch_restart': 'yes',
             'primary_switch_disable_archive_restore': 'yes',
             'close_detached_after': 300,
+            # How long to wait for walreceiver to actually stop after emptying primary_conninfo + reload.
+            # Startup applies SIGHUP asynchronously; if it does not stop the receiver in time, abort failover.
+            'walreceiver_disable_timeout': 5,
         },
         'commands': {
             'promote': '/usr/lib/postgresql/10/bin/pg_ctl promote -D %p',
@@ -126,7 +129,7 @@ def read_config(filename=None, options=None):
         },
         'debug': {
             'election_loser_timeout': 0,  # Timeout for election losers. For test purposes only.
-            'pre_pause_sleep': 0,  # Sleep before pg_wal_replay_pause() in _can_do_failover. For test purposes only.
+            'pre_pause_sleep': 0,  # Sleep before disabling walreceiver in _can_do_failover. For test purposes only.
             'election_lsn_read_sleep': 0,  # Sleep right after reading wal_receive_lsn for the election vote. For test purposes only.
         },
     }
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -126,6 +126,8 @@ def read_config(filename=None, options=None):
         },
         'debug': {
             'election_loser_timeout': 0,  # Timeout for election losers. For test purposes only.
+            'pre_pause_sleep': 0,  # Sleep before pg_wal_replay_pause() in _can_do_failover. For test purposes only.
+            'election_lsn_read_sleep': 0,  # Sleep right after reading wal_receive_lsn for the election vote. For test purposes only.
         },
     }
 
--- a/src/main.py
+++ b/src/main.py
@@ -275,7 +275,7 @@ class pgconsul(object):
             primary_fqdn = db_state.get('primary_fqdn')
             if current_primary is None and primary_fqdn is not None:
                 self.zk.write_maintenance_primary(primary_fqdn)
-        elif maintenance_status == 'disable':
+        elif maintenance_status in ('disable', ''):
             # maintenance node exists with 'disable' value, we are not in maintenance now
             # and should delete this node. We delete it recursively, we don't won't to wait
             # all cluster members to delete each own node, because some of them may be
--- a/src/main.py
+++ b/src/main.py
@@ -275,7 +275,7 @@ class pgconsul(object):
             primary_fqdn = db_state.get('primary_fqdn')
             if current_primary is None and primary_fqdn is not None:
                 self.zk.write_maintenance_primary(primary_fqdn)
-        elif maintenance_status in ('disable', ''):
+        elif maintenance_status == 'disable':
             # maintenance node exists with 'disable' value, we are not in maintenance now
             # and should delete this node. We delete it recursively, we don't won't to wait
             # all cluster members to delete each own node, because some of them may be
--- a/src/main.py
+++ b/src/main.py
@@ -1588,6 +1588,11 @@ class pgconsul(object):
             logging.warning('Promote is not allowed with given configuration.')
             return False
 
+        pre_pause_sleep = self.config.getfloat('debug', 'pre_pause_sleep', fallback=0)
+        if pre_pause_sleep:
+            logging.debug('Sleep for test purposes before pg_wal_replay_pause: %s', pre_pause_sleep)
+            time.sleep(pre_pause_sleep)
+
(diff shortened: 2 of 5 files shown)


munakoiso:
For now there is just simple fix (which does not work in specific case) and test, that should fails

munakoiso:
https://github.com/yandex/pgconsul/actions/runs/30755545167/job/91516999984?pr=199#step:5:5212 - failed run

munakoiso:
green test - https://github.com/yandex/pgconsul/actions/runs/30756758202/job/91606118373
but maintenance tests is too flaky, i'll fix it

src/__init__.py
```diff
@@ -108,6 +108,9 @@ def read_config(filename=None, options=None):
             'primary_switch_restart': 'yes',
             'primary_switch_disable_archive_restore': 'yes',
             'close_detached_after': 300,
+            # How long to wait for walreceiver to actually stop after emptying primary_conninfo + reload.
+            # Startup applies SIGHUP asynchronously; if it does not stop the receiver in time, abort failover.
+            'walreceiver_disable_timeout': 5,
```
vicpopov, line 113: Maybe a less-specific `setting_apply_timeout`?
We recently discovered that `global/postgres_timeout` is only used in waiting for `restore_command` setting change, so it is not so good named too...

And with this we can encapsulate timeout into setting-changing methods, and keep code cleaner.

src/main.py
```diff
@@ -275,7 +275,7 @@ def update_maintenance_status(self, db_state, zk_state):
             primary_fqdn = db_state.get('primary_fqdn')
             if current_primary is None and primary_fqdn is not None:
                 self.zk.write_maintenance_primary(primary_fqdn)
-        elif maintenance_status == 'disable':
+        elif maintenance_status in ('disable', ''):
```
vicpopov: Comment at https://github.com/yandex/pgconsul/pull/199/changes#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R240 says it can be `None`, but not empty string.

src/pg.py
```diff
@@ -863,21 +863,6 @@ def is_replaying_wal(self, check_time):
         replay_diff = self.get_replay_diff()
         return prev_replay_diff < replay_diff
 
-    def pg_wal_replay_pause(self) -> bool:
```
vicpopov, line 868: I'm not sure we should abandon replay pause entirely.
We must promote a replica that both:
- has the latest WAL from primary
- has replayed all WAL up to it
- 
To ensure replay, we have resume replay later: https://github.com/yandex/pgconsul/pull/199/changes#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R1674
But you are removing pausing.
It is at least inconsistent, and I would suggest to dig here, if it can lead to promote of uncaught-up replica.

tests/features/failover_with_network_inconsistency.feature
```diff
@@ -75,3 +75,89 @@ Feature: Failover with network inconsistency
...
+        # Wait until both replicas have entered _can_do_failover and are sleeping
+        # right before disabling walreceiver
+        Then container "postgresql2" pgconsul log contains "Sleep for test purposes before disabling walreceiver"
+        Then container "postgresql3" pgconsul log contains "Sleep for test purposes before disabling walreceiver"
+        # Old primary comes back before either replica has disabled its walreceiver
+        When we start "postgres" in container "postgresql1"
+        Then container "postgresql2" walreceiver is streaming from container "postgresql1"
+        Then container "postgresql3" walreceiver is streaming from container "postgresql1"
+        # Freeze startup for a bounded window (auto-CONT after N seconds).
+        # On current code replicas vote while startup is still frozen and walreceiver
```
vicpopov: What is `current code`? It won't be current after we merge this PR.
Maybe `code before <data>`?

tests/steps/cluster.py
```diff
@@ -1512,6 +1514,101 @@ def step_container_pause_replaying_wal(context, name):
...
+
+    The deadline is enforced in the test process (not via Postgres
+    statement_timeout): run the query in a helper thread, and on timeout
+    cancel the backend via the connection and treat that as success.
+    """
+    container = _get_container(context, name)
+    timeout_sec = int(timeout_ms) / 1000.0
+    conn = psycopg2.connect(
+        host=helpers.container_get_host(),
+        port=helpers.container_get_tcp_port(container, 5432),
```
vicpopov, line 1595: Why `5432`? User-level load by convention connects through pooler, `6432`.

munakoiso: We cannot rely on pooler for data security. We can only rely on pg mechanisms.

src/main.py
munakoiso: This change will be merged in pull/188. 
Reverted here

src/__init__.py
munakoiso: this timeout not for "set variable", but for "wait til receiver will be disabled". Not the same as  setting_apply_timeout

src/pg.py
munakoiso: Pause not give us any guarantee (receiver with sync_commit=on will continue to receive wals and sends ack to an old master), but increases replay lag before resume

src/main.py
```diff
@@ -1590,22 +1590,35 @@ def _can_do_failover(self, switchover_in_progress=False):
             logging.warning('Promote is not allowed with given configuration.')
             return False
 
-        if not self.db.pg_wal_replay_pause():
+        pre_pause_sleep = self.config.getfloat('debug', 'pre_pause_sleep', fallback=0)
```
vicpopov, line 1593: `pre_pause_sleep` - setting name is about pause, but we removed pause from our code, we are now disabling walreceiver, we should rename this debug setting.
But NIT since it is a debug-only setting.

vicpopov (approved):

src/main.py
munakoiso: in separate pr

PR #200, opened 2026-08-03
MDB-48301: Wait for walreceiver stop before failover election (stable branch)

munakoiso:
## Problem

During failover a replica clears `primary_conninfo` and reloads so that it stops receiving WAL from the old primary before it reads LSN and votes. That disable is applied by Postgres **startup asynchronously**. Reload returning does **not** mean walreceiver is already gone.

Until startup actually stops the receiver, the replica can still:
- receive WAL onto disk;
- send sync acknowledgements to the old primary.

`pg_wal_replay_pause()` does not close this window: it only stops **replay**, not receive/ack. So a replica can vote with an LSN taken while walreceiver is still alive, and the resurrected old primary can still get sync commits acknowledged after voting has started — a split-brain / divergent-write risk.

## Fix

Before entering the election:
1. set `primary_conninfo` to empty and reload (if needed);
2. wait until `pg_stat_wal_receiver` is empty, up to `replica.walreceiver_disable_timeout` (default 5s);
3. if the receiver does not stop in time — abort this failover attempt;
4. only then read LSN and participate in the election.

`pg_wal_replay_pause` was removed from this path as ineffective for the problem above.

## Test

New scenario in `failover_with_network_inconsistency.feature` forces the worst case:
- replicas enter failover and hit a debug sleep before disable;
- old primary is started again and walreceivers reconnect;
- replica `postgres: startup` is SIGSTOP’d (auto-CONT after N seconds), so disable cannot take effect while frozen;
- after votes appear in ZK, a sync `CREATE TABLE` on the old primary must **not** complete (app-side deadline; `statement_timeout` does not cancel SyncRep waits).

With the fix, replicas do not vote until the receiver is really stopped (after CONT / successful wait). Without the wait, they can vote while the receiver is still acking and the write can succeed.

Debug-only knobs for the scenario: `debug.pre_pause_sleep`, `debug.election_lsn_read_sleep`.

New scenario green with this change; red if wait-for-stop is removed

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -106,6 +106,9 @@ def read_config(filename=None, options=None):
             'primary_switch_restart': 'yes',
             'primary_switch_disable_archive_restore': 'yes',
             'close_detached_after': 300,
+            # How long to wait for walreceiver to actually stop after emptying primary_conninfo + reload.
+            # Startup applies SIGHUP asynchronously; if it does not stop the receiver in time, abort failover.
+            'walreceiver_disable_timeout': 5,
         },
         'commands': {
             'promote': '/usr/lib/postgresql/10/bin/pg_ctl promote -D %p',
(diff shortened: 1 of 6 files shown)


munakoiso:
failed run before fix - https://github.com/yandex/pgconsul/actions/runs/30801016722/job/91649194602?pr=200

munakoiso:
green test - https://github.com/yandex/pgconsul/actions/runs/30979563616/job/92221296133?pr=200

aak74 (approved):
LGTM

PR #206, opened 2026-08-05
MDB-41951: consolidate factory modules and add documents ADR-0003/0004

aak74:
- Add ADR-0003: formalize two-layer ZK architecture (ZkClient transport vs
  Zookeeper domain) with exception translation contract
- Add ADR-0004: establish Factory + Config-Builder convention for
  infrastructure components
- Merge replication_manager_factory.py into replication_manager.py following
ADR-0004 factory convention.


src/pg.py
```diff
@@ -952,5 +954,61 @@ def check_extension_installed(self, name):
         result = cur.fetchall()
         return len(result) == 1
 
+    def is_host_reachable(self, primary: str | None = None, check_primary: bool = True) -> bool:
+        """
+        Check if a host is accessible via the postgres protocol.
+
+        Returns False if the host is reachable (alive), True if it is not.
```
munakoiso: "is_host_reachable" "Returns False if the host is reachable"

We need to invert return here and in function calls (or rename function)

munakoiso: also this one looks strange - no primary = return false = primary available
'''
if not primary:
            primary = self.get_primary_fqdn()
            if not primary:
                return False
'''

aak74: fixed

munakoiso (approved):

PR #214, opened 2026-08-25
Make failover and switchover blocking with local state

munakoiso:
- make failover and switchover blocking operations
- move host-local operation state from ZooKeeper to crash-safe local files
- separate failover and switchover metadata
- handle primary lock cleanup after failed promotion
- simplify state-machine routing and cleanup
- store switchover state in one json, use write over exact zk-node version (aka CAS) in switchover to prevent ABA problem
- update unit tests for the new behavior


aak74 (approved):
LGTM

PR #217, opened 2026-09-06
MDB-47163: option to keep the primary out of the commit quorum

vicpopov:
- `quorum_includes_primary` (default `yes`, current behaviour) counts the primary's own copy of a commit towards the quorum. With `no` the quorum is a majority of the replicas alone, so a cluster degraded to three hosts keeps `ANY 2(...)` instead of falling back to `ANY 1(...)`, and survives the loss of its primary together with one replica.
- Promote safety is decided by the quorum size the primary actually required, read from its `synchronous_standby_names` in ZK, and falls back to the size the default configuration installs unless that value describes this very group of hosts. Deriving the size from the local option would let a host promote with commits missing while the option is being rolled out over a cluster.
- Unit tests pin the quorum size, the promote threshold and every way into its fallback; the behave feature pins the quorum size for clusters of 1 to 5 hosts and both failover outcomes on a three host cluster that loses its primary and one replica at once.


src/helpers.py
```diff
@@ -170,6 +170,17 @@ def make_current_replics_quorum(replics_info: ReplicaInfos, alive_hosts):
     return {host for host, app_name in alive_hosts_map.items() if app_name in alive_replics}
 
 
+def quorum_from_ssn(ssn):
```
vicpopov: `quorum_size_from_ssn`? quorum itself is enumeration of replics

src/replication_manager.py
```diff
@@ -408,8 +412,20 @@ def update_replication_type(self, db_state, ha_replics, set_quorum_to=None):
                 self._zk.write(self._zk.QUORUM_PATH, quorum_hosts, preproc=json.dumps)
                 logging.info('Turned synchronous replication ON.')
 
+    def _quorum_size(self, replicas_number):
+        """
+        Number of replicas that have to confirm a commit.
+
+        The primary holds every commit it confirms, so counting it in makes the
+        quorum a majority of the whole cluster. Leaving it out makes the quorum a
+        majority of the replicas alone, so a commit outlives the primary.
```
vicpopov: A commit outlives the primary in the first case too because we promote the replica that wins elections, and thus contain all commits the primary has confirmed. Something more accurate should be here.

src/helpers.py
vicpopov: Please annotate return value type, it is not clearly visible that this function returns tuple.

tests/features/quorum_includes_primary.feature
```diff
@@ -0,0 +1,158 @@
...
+
+    Examples: quorum of the whole cluster, the primary included
+        | includes_primary | hosts | replicas | quorum |
+        | yes              | 5     | 4        | 2      |
+        | yes              | 4     | 3        | 2      |
+        | yes              | 3     | 2        | 1      |
+        | yes              | 2     | 1        | 1      |
+        | yes              | 1     | 0        | 0      |
+
+    Examples: quorum of the replicas alone, one more replica on odd-sized clusters
```
vicpopov: `one more replica required for quorum on odd-sized clusters`?

tests/steps/cluster.py
```diff
@@ -736,6 +773,12 @@ def step_disconnect_container(context, name):
         context.networks[netname].disconnect(container)
 
 
+@when('we disconnect from network containers "(?P<names>[,a-zA-Z0-9_-]+)"')
+def step_disconnect_containers(context, names):
+    for name in names.split(','):
+        step_disconnect_container(context, name)
+
+
```
vicpopov: We use this step only twice, so it would be a total of 4 lines with "we disconnect from network container", but introducing this step adds more lines of code.
Can we please just write it plainly that in scenario?

src/replication_manager.py
```diff
@@ -449,11 +469,44 @@ def is_promote_safe(self, host_group, replica_infos: ReplicaInfos):
...
+
+        Computing it from the local `quorum_includes_primary` would be wrong: while the
+        option is being rolled out the hosts disagree about it, and a size larger than
+        the primary's would let us promote a host that misses confirmed commits. Unless
+        the primary's value describes this very group of hosts we fall back to the size
+        the default configuration installs, which is the stricter of the two the option
+        can produce and the one this check used before the option existed.
+        """
+        last_primary = self._zk.noexcept_get(self._zk.LAST_PRIMARY_PATH)
+        primary_ssn = self._zk.noexcept_get(self._zk.get_ssn_value_path(last_primary)) if last_primary else None
```
vicpopov: Мне кажется, вот эти обращения к ZK следует вынести в метод, вызывающий этот - в `is_promote_safe`, так как там у нас уже есть чтение из ZK, лучше их сгруппировать там, а эту функцию оставить чисто вычислительной.

src/replication_manager.py
```diff
@@ -449,11 +469,44 @@ def is_promote_safe(self, host_group, replica_infos: ReplicaInfos):
...
+    def _primary_quorum_size(self, sync_quorum):
+        """
+        Quorum size the primary required, taken from its own synchronous_standby_names.
+
+        Computing it from the local `quorum_includes_primary` would be wrong: while the
+        option is being rolled out the hosts disagree about it, and a size larger than
+        the primary's would let us promote a host that misses confirmed commits. Unless
+        the primary's value describes this very group of hosts we fall back to the size
+        the default configuration installs, which is the stricter of the two the option
+        can produce and the one this check used before the option existed.
```
vicpopov: Это кажется излишне многословным объяснением.
Можно короче, в духе: локальное вычисление с основой на опцию `quorum_includes_primary` может быть некорректным в процессе выкатки.

vicpopov:
Another variant in #230, which is an extension ot this one, was accepted

PR #219, opened 2026-09-09
Stabilize faultstorm test infrastructure

munakoiso:
Fixes faultstorm setup and feature isolation.

- remove the obsolete PostgreSQL setting from the test image
- install faultstorm with current pip syntax
- clean conflicting compose networks before faultstorm startup
- make Behave skip @skip and propagate failures
- wait for normal cluster recovery between faultstorm features


aak74 (approved):
Approval to run tests: Jepsen and Faultstorm.
We will address the issues related to this PR in separate pull requests.

PR #229, opened 2026-09-12
MDB-47163: forced quorum replicas number (stable branch)

vicpopov:
- `forced_quorum_replicas_number` sets how many replicas have to confirm a commit instead of the majority pgconsul computes on its own, never more than the group the confirmations are required from. A three host cluster then keeps `ANY 2(...)` and survives the loss of its primary with one replica.
- Promote safety is decided by the quorum size the primary actually required, read from its `synchronous_standby_names` in ZK. When that value is unusable the whole quorum group has to be alive: an administrative size can be any number, so no lower threshold is safe while the setting is rolled out.
- A changed quorum size now reaches PostgreSQL without a change of the quorum membership, and the quorum is parsed with the spacing PostgreSQL keeps in a hand-set value: the first left the setting unapplied, the second made a ZK loss close the pooler.
- Unit tests pin the quorum size, its cap, the promote threshold and both of those fixes; the behave feature pins the quorum size for clusters of 1 to 5 hosts and both failover outcomes on a three host cluster that loses its primary and one replica at once.


src/replication_manager.py
```diff
@@ -449,11 +473,39 @@ def is_promote_safe(self, host_group, replica_infos: ReplicaInfos):
...
+        quorum = helpers.quorum_from_ssn(primary_ssn)
+        if quorum is not None:
+            quorum_size, app_names = quorum
+            group_matches = set(app_names) == {helpers.app_name_from_fqdn(host) for host in sync_quorum}
+            if group_matches and 1 <= quorum_size <= len(sync_quorum):
+                logging.info('Primary %s required %s confirmations of %s.', last_primary, quorum_size, sync_quorum)
+                return quorum_size
+        logging.warning(
+            'Primary %s has no synchronous_standby_names in ZK describing the quorum %s (it is %s), '
+            'falling back to a quorum of one, which demands the whole group alive.',
```
vicpopov: `which demands the whole group alive.` немного недосказано получается, неясно зачем это требуется.
Но это уже концепция уровнем выше, отсюда логгировать это не совсем логично.

src/replication_manager.py
```diff
@@ -449,11 +473,39 @@ def is_promote_safe(self, host_group, replica_infos: ReplicaInfos):
...
+            last_primary,
+            sync_quorum,
+            primary_ssn,
+        )
+        return 1
```
vicpopov: А почему 1? Если в кворуме нет хостов то должно быть 0.
И вроде бы это возможный вариант - если наша реплика не успела войти в кворум, когда случился отказ.

vicpopov:
Accepted other variant in #230

PR #231, opened 2026-09-12
MDB-47163: do not publish the quorum before the promote succeeds

vicpopov:
- `_do_failover` installs the pre-promote `synchronous_standby_names` and writes its quorum to ZK in one step, while `last_leader` already names the candidate. A promote that then fails leaves ZK advertising a quorum that governed no commit, written by a host that never became primary.
- Today that only misstates the group a later failover checks itself against; it turns into a data-loss path as soon as the quorum size is read back from ZK. The quorum is now published by the primary that holds the lock, on its own next iteration, and a unit test pins that the pre-promote path writes only PostgreSQL.


src/replication_manager.py
```diff
@@ -372,8 +372,10 @@ def update_replication_type(self, db_state, ha_replics, set_quorum_to=None):
         current = self._db.get_replication_state()
         logging.info('Current replication type is %s.', current)
         if set_quorum_to is not None:
+            # The quorum stays unpublished until the promote that follows succeeds: a
+            # quorum written for a host that never became primary describes commits that
+            # were never made under it.
             if self.change_replication_to_quorum(set_quorum_to):
-                self._zk.write(self._zk.QUORUM_PATH, set_quorum_to, preproc=json.dumps)
```
vicpopov, line 376: На самом деле, тут метод `update_replication_type`, использующийся только 2 раза в коде, не должен быть одним методом: тут совершенно отдельные пути выполнения.
Это должно быть отдельным методом специально применяемым при failover.

munakoiso (approved):

